Issue #75, opened 2024-08-19
Titles in collection should be sorted by release date


yihaolee85:
**Describe the bug**
When selecting a collection, the titles are sorted alphabetically. They should be sorted by release date similar to what is in the Jellyfin web UI.
Currently collections like James Bond, Pirates of the Caribbean, Mission Impossible, Harry Potter will all be show out of order.

**To Reproduce**
1. Go to Library
2. Click on Collections
3. Select any collection
4. Titles are not sorted by release order

**Screenshots**
![IMG_1091](https://github.com/user-attachments/assets/217156d0-78f0-4e81-b70f-ca0c852cf555)

**Smartphone (please complete the following information):**
 - Device: iPhone XR
 - OS: iOS 17.6.1
 - Version 0.6.1 (4)

yihaolee85:
@fredrikburmester thank you for the fix, downloaded 0.6.2 (3) and noticed a few things:

1) Titles are now sorted descending, they should be ascending to see the titles in order
2) Based on the commit, looks like you are using ProductionYear. I suggest using PremiereDate instead for sorting, if possible, as there can be 2 titles released in the same year in a collection (example Matrix Reloaded and Matrix Revolutions were both released in 2003)
3) This sorting has spilled over to the other views when switch back (e.g. sort Movies by Name, go into collections, select a collection, go back to Movies and now Movies is sorted by release date instead). I'll probably raise a new issue for this with better description

Thanks!

fredrikburmester:
Thank you for this detailed testing, I really appreciate it! I will look into all the things you mentioned.

PR #85, opened 2024-08-21
Update [collectionId].tsx to fix sort order

yihaolee85:
Update [collectionId].tsx to fix sort order #75


fredrikburmester:
Thank you!

PR #129, opened 2024-09-15
removes the need for http/https in the url bar

Simon-Eklundh:
- [X] working with http domains
- [X] working with https domains
- [X] working with local ips
- [ ] working with remote ips  [ UNTESTED but should work logically ]
- [x] working with tailscale [ UNTESTED ]


app/login.tsx
```diff
@@ -78,13 +80,40 @@ const Login: React.FC = () => {
...
     }
   };
+  async function checkUrl(url: string) {
+    // remove trailing / so we don't double add double / in my checker
+    if (url.endsWith("/")) {
+      url = url.slice(0, -1);
+    }
 
-  const handleConnect = (url: string) => {
+    if( await fetch("https://" + url + "/web/#/home.html", {mode: 'cors'}).then((response) => {
```
sjorge, line 89: Would it not be better to hit the server info API endpoint? It's un authenticated and also works for jellyfin installations without the webui.

Simon-Eklundh: genuinely just didn't think of the /System/Info/Public endpoint being a thing :laughing: 

/System/Info/Public seems to work, thanks for the comment!

fredrikburmester:
Should we add some kind of indication that the app will auto-fill the scheme? Maybe a text or placeholder or other nice UI design?

Simon-Eklundh:
> Should we add some kind of indication that the app will auto-fill the scheme? Maybe a text or placeholder or other nice UI design?

I did consider it. The original has something along the lines of
"The url needs to start with http/https"
So I did consider
"The url should not need to start with http/https" but i didn't like it.

Something like "if you don't add http/Https we will add it for you" but idk, I'm just kinda stuck in the "I have no idea what to write as UI"-state...

Simon-Eklundh:
> Should we add some kind of indication that the app will auto-fill the scheme? Maybe a text or placeholder or other nice UI design?

I think having a grey http/Https text at the front would probably be enough...

Although if for some reason there's some big thing I've missed that might confuse someone who needs to write it anyhow (not that I think someone would need to)

fredrikburmester:
I improved the code with timeouts, docs string, included checks even if the user entered http/s in the string, improved alert message, added loader and disabled button during check. Can you please check and test the code @Simon-Eklundh ?

I also tested with Tailscale ips, external domain and local ips. All worked.

Simon-Eklundh:
> I improved the code with timeouts, docs string, included checks even if the user entered http/s in the string, improved alert message, added loader and disabled button during check. Can you please check and test the code @Simon-Eklundh ?
> 
> I also tested with Tailscale ips, external domain and local ips. All worked.

will do!

Simon-Eklundh:
@fredrikburmester  Looks good to me!

(took me a hot minute to realize why you wrote ```new URL(url).host```)

PR #151, opened 2024-10-02
Added support for more libraries to use the recently added section

Alexk2309:
Allows for more libraries to be seen in the recently added section. When I was using the app I wasn't able to see my most recently viewed Anime, this feature should fix that

BEFORE:
image

AFTER:

image


Simon-Eklundh:
I like it, but I'd love having the ability to pick which to show in some way (I have two libraries for backups which I don't really want to show :)

not saying you have to do it but it is something I'd personally like :)

Alexk2309:
> I like it, but I'd love having the ability to pick which to show in some way (I have two libraries for backups which I don't really want to show :)
I believe you can not enable certain libraries for each user using jellyfin, and it wouldn't show. Then this shouldn't pop up. I don't know if this should be done using the app lol.

PR #181, opened 2024-10-12
Update visual playback when exiting video

Alexk2309:
This PR addresses an issue where the frontend UI was not updating instantly when a video stops playing. The changes ensure that the relevant queries are invalidated correctly, triggering an immediate UI update.

Note:

The new lines were just copied from the PlayedStatus.tsx file, this code could be refactored so that way both these two components could use one function. Please add a comment if you would like this.

## Summary by Sourcery

Bug Fixes:
- Ensure the frontend UI updates instantly when a video stops playing by invalidating relevant queries.

### DIFF
--- a/app/(auth)/play-video.tsx
+++ b/app/(auth)/play-video.tsx
@@ -131,15 +131,12 @@ export default function page() {
       queryKey: ["seasons"],
     });
     queryClient.invalidateQueries({
-      queryKey: ["nextUp-all"],
-    });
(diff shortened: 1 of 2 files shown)


fredrikburmester:
Could you please remove the duplicate 'nextUp-all' query and add the queryClient to the dep array and i'll merge this! Thank you!

Alexk2309:
@fredrikburmester Okay changes made!

PR #210, opened 2024-10-29
Add MacOS fullscreen support

MateuszKukiela:
**Summary:**
Added support for macOS to use true screen size with `UISupportsTrueScreenSizeOnMac: true`. This enables the app to fully utilize custom resolutions and fullscreen mode on macOS, enhancing the viewing experience for users. 

**Change:**
- Set `UISupportsTrueScreenSizeOnMac` to `true` to enable native fullscreen and resolution support on macOS.

**Testing:**
Tested on an M1 Mac to confirm the app utilizes full-screen mode and native resolutions as expected on macOS.


fredrikburmester:
Hey thanks for this! Will merge as soon as i've tested.

PR #301, opened 2024-12-22
Changed to only add safe areas on top

Alexk2309:
## Summary by Sourcery

New Features:
- Removed left and right safe area insets for the top controls to provide a full-width experience.


fredrikburmester:
Please make this an option in the settings. Make sure the default behavior is to have the safe areas **enabled** for all sides. @Alexk2309

fredrikburmester:
I implemented safe areas as a setting.

PR #343, opened 2025-01-01
feat: add favorites functionality with FavoriteButton component and l…

sldless:


bun.lockb
lostb1t, line 1: this should not be deleted

Issue #361, opened 2025-01-04
[Bug]: Playback always in portrait mode


anshuman852:
### What happened?

Basically the title, i just installed the app, came here from reddit after reading about the jellyseer integration. love the app ui so far. but the problem is that it always plays in portrait mode, 
Can the default behaviour be made, like
Once you press play the video plays in landscape (like in mxplayer, youtube, jellyfin app, and TON of other apps)
And then you can only switch between the two landscape orientation modes, landscape left or right 

### Reproduction steps

1. play anything on streamyfin
2. it runs in portrait
3. 

### Which device and operating system are you using?

Samsung s23 5g , one ui 6.1

### Version

0.24.0

### If applicable, please add screenshots to help explain your problem. You can drag and drop images here or paste them directly into the comment box.

![image](https://github.com/user-attachments/assets/b765030d-80e5-454b-8a5b-ce6fbcc91057)

Ryan0204:
I guess you can set it on Settings > Other > Auto rotate (disable) > Video orientation?

anshuman852:
> I guess you can set it on Settings > Other > Auto rotate (disable) > Video orientation?  

Yes i know, but the default behaviour is odd tbh, unlike anything i have ever seen

fredrikburmester:
Fixed with #383

PR #383, opened 2025-01-06
Change ScreenOrientation to landscape right by default and added toggleSafeArea for all videos

Ryan0204:
re: #361 
Netflix or others major streaming app open videos in landscape mode by default as well.
I think it's more user-friendly.


Ryan0204:
[c86a843](https://github.com/streamyfin/streamyfin/pull/383/commits/c86a8438e56bd040b0f33aebe2e83f29244ac9da) Allow all videos accessible to the toggleSafeArea button. Allowing iPhone with dynamic island to watch videos without any part of the videos covered.

fredrikburmester:
This is awesome! Though we need to add rotate back on exit player. Could you add that?

Ryan0204:
I've added it! Please have a look, thanks!

Issue #390, opened 2025-01-06
[Feature Request] Better media controls and behavior


7eventy7:
First of all I want to say you all are awesome for working on these features, bugs and other changes so quick, thank you!

I have two main suggestions:

1. Currently, whenever something is playing and you tap the screen to view the play controls, they will remain there forever until you tap again which causes them to hide. It would be nice if there was like a cooldown where the controls automatically went away after not interacting with the screen for 10-15 seconds. Maybe even make the duration a setting, although that seems a bit unneeded.

2. The gesture controls for brightness and volume can currently only be changed when the media controls are opened with a tap (mentioned above). It would be really nice if you could use these gestures outside of the controls menu, swiping on the right or left side of the screen to increase/decease.

7eventy7:
Still relevant!

PR #394, opened 2025-01-08
enhancement: auto hide control after 5 seconds

Ryan0204:
#390 
1. Added auto hide control after 5 seconds


fredrikburmester:
This is great but we need to handle when the user is scrubbing/seeking. Right now the controls disappear even if the user is actually swiping on the screen and seeking. Could you fix this? Thanks!

Ryan0204:
@fredrikburmester I have fixed it but I believe there are simpler code than mine haha.

Issue #401, opened 2025-01-09
Streamyfin MAM Plugin solution


herrrta:
**Describe the solution you'd like**
Allow admins to use [jellyfin-plugin-streamyfin](https://github.com/streamyfin/jellyfin-plugin-streamyfin) to configure Streamyfin for their users.

We can offer as much settings as we can so that admins can create a unique Streamyfin app experience for their users.

Potentially blacklist/whitelist users from receiving these forced settings?

**Additional context**
Will be nice since some of us offer use of our jellyfin instance to family... 
and well sometimes family doesnt know how to navigate to settings and configure the application for an optimized experience....

herrrta:
initial changes to support our settings as a configuration in this pr
https://github.com/streamyfin/jellyfin-plugin-streamyfin/pull/1

PR #402, opened 2025-01-09
Streamyfin MAM Plugin solution

herrrta:
implements #401

~**DONT MERGE YET**~

## Summary by Sourcery

New Features:
- Added the ability for admins to lock settings through the Streamyfin MAM plugin.

### DIFF
--- a/app/(auth)/(tabs)/(home)/downloads/index.tsx
+++ b/app/(auth)/(tabs)/(home)/downloads/index.tsx
@@ -4,7 +4,7 @@ import { MovieCard } from "@/components/downloads/MovieCard";
 import { SeriesCard } from "@/components/downloads/SeriesCard";
 import { DownloadedItem, useDownload } from "@/providers/DownloadProvider";
 import { queueAtom } from "@/utils/atoms/queue";
-import { useSettings } from "@/utils/atoms/settings";
+import {DownloadMethod, useSettings} from "@/utils/atoms/settings";
 import { Ionicons } from "@expo/vector-icons";
 import { useNavigation, useRouter } from "expo-router";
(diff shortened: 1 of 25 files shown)


herrrta:
DO NOT MERGE this is to track changes

This commit is only partially using the centralized plugin solution to login into Jellyseerr the same time a user logs into Jellyfin.


TODO:
- Use plugin settings to determine which app settings are locked from modification

lostb1t:
@herrrta you can make it a draft pr if it's not ready yet

herrrta:
We should also setup some polling for this configuration on the app itself.

Just so we can make an attempt to always be in "sync" with what the admin dictates for their users

Stetsed:
> We should also setup some polling for this configuration on the app itself.
> 
> Just so we can make an attempt to always be in "sync" with what the admin dictates for their users
> 
> If we can do this with ws/s that would be better but am unsure if this is something we'd have to do in plugin side

Fyi, because this is a simple JSON request I think it would make most sense to just tie it to another standard action. Like for example when the app is opened or something similar. Having this be a constant polling task would very much be overkill imho.

herrrta:
> > We should also setup some polling for this configuration on the app itself.
> 
> > 
> 
> > Just so we can make an attempt to always be in "sync" with what the admin dictates for their users
> 
> > 
> 
> > If we can do this with ws/s that would be better but am unsure if this is something we'd have to do in plugin side
> 
> 
> 
> Fyi, because this is a simple JSON request I think it would make most sense to just tie it to another standard action. Like for example when the app is opened or something similar. Having this be a constant polling task would very much be overkill imho.

I want as live as i can get. Even if its 1,2,5 intervals

PR #403, opened 2025-01-09
Feature: Remove duplicate names from media sources

lostb1t:
Before:

Screenshot 2025-01-09 at 10 15 55

after:

Screenshot 2025-01-09 at 10 13 13

## Summary by Sourcery

New Features:
- Display shorter media source names in the dropdown.


lostb1t:
i removed the bitrate because its already shown under video details

Issue #407, opened 2025-01-09
[Bug]: keyboard always open on app launch


Panda260:
### What happened?

When i open the App my keyboard opens.
but this only hapens if the app is not running in the background when it is running in the background it is fine
when i clear the cache/the open apps in the background the keyboard will open again

### Reproduction steps

Open the App

### Which device and operating system are you using?

Xiaomi 14T Pro
Xiaomi HyperOS 1.0.14.0
Android 14
StreamyFin App Version **0.24** not 0.25 github is wrong

### Version

0.25.0

Panda260:
maybe thats it because the login screen opens for a short second https://workupload.com/file/rjsV6CrNWV9

whoopsi-daisy:
> ### What happened?
> 
> When i open the App my keyboard opens. but this only hapens if the app is not running in the background when it is running in the background it is fine when i clear the cache/the open apps in the background the keyboard will open again
> ### Reproduction steps
> 
> Open the App
> ### Which device and operating system are you using?
> 
> Xiaomi 14T Pro Xiaomi HyperOS 1.0.14.0 Android 14 StreamyFin App Version **0.24** not 0.25 github is wrong
> ### Version
> 
> 0.25.0
> ### If applicable, please add screenshots to help explain your problem. You can drag and drop images here or paste them directly into the comment box.
> 
> _No response_

Same behavior on HyperOS 2 / Android 15

Panda260:
> > ### What happened?
> > 
> > When i open the App my keyboard opens. but this only hapens if the app is not running in the background when it is running in the background it is fine when i clear the cache/the open apps in the background the keyboard will open again
> > ### Reproduction steps
> > 
> > Open the App
> > ### Which device and operating system are you using?
> > 
> > Xiaomi 14T Pro Xiaomi HyperOS 1.0.14.0 Android 14 StreamyFin App Version **0.24** not 0.25 github is wrong
> > ### Version
> > 
> > 0.25.0
> > ### If applicable, please add screenshots to help explain your problem. You can drag and drop images here or paste them directly into the comment box.
> > 
> > _No response_
> 
> Same behavior on HyperOS 2 / Android 15

What device?

MulverineX:
Same behavior on Stock ROM, Google Pixel 9 Pro, Android 14

fredrikburmester:
Fixed with #437

MulverineX:
(not actually fixed)

Panda260:
nope not fixed

[...]

PR #434, opened 2025-01-15
fix: Keyboard breaking UI when already logged in

tom-heidenreich:
> Note: This may also be an issue on ios, but I could only test on android

When opening the app the login page shows up for a split second and because of the autoFocus on the username input, the keyboard shows up, which could mess up to general ui (happened like half of the time).

The fix is a custom property `autoFocusDelay` on the Input component, which triggers the autoFocus after a given amount of time.
The username input uses this `autoFocusDelay` instead of `autoFocus` with a short time, so the keyboard does not show up, when already logged in, but not too long, so it isn't noticeable when not logged in.

## Summary by Sourcery

Bug Fixes:
- Fixed an issue where the keyboard would appear briefly on the login screen when the app was opened while already logged in.

### DIFF
--- a/app/login.tsx
+++ b/app/login.tsx
@@ -223,7 +223,8 @@ const Login: React.FC = () => {
                       setCredentials({ ...credentials, username: text })
                     }
                     value={credentials.username}
-                    autoFocus
+                    // use auto focus with small delay so keyboard only shows up when not logged in
+                    autoFocusDelay={500}
                     secureTextEntry={false}
                     keyboardType="default"
                     returnKeyType="done"
--- a/components/common/Input.tsx
+++ b/components/common/Input.tsx
@@ -13,13 +13,13 @@ export function Input(props: TextInputProps & AutoFocusDelayProps) {
   const { style, ...otherProps } = props;
   const inputRef = React.useRef<TextInput>(null);
 
-  if(props.autoFocus && props.autoFocusDelay) {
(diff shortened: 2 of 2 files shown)


tom-heidenreich:
This fixes #407

tom-heidenreich:
Yeah, there should probably be a loading page. Often other apps show a page with just the logo. Could be an option

tom-heidenreich:
You could not render the login page like this
```diff
--- a/providers/JellyfinProvider.tsx
+++ b/providers/JellyfinProvider.tsx
@@ -338,11 +338,12 @@ export const JellyfinProvider: React.FC<{ children: ReactNode }> = ({
     initiateQuickConnect,
   };
 
-  useProtectedRoute(user, isLoading || isFetching);
+  let isLoadingOrFetching = isLoading || isFetching;
+  useProtectedRoute(user, isLoadingOrFetching);
 
   return (
     <JellyfinContext.Provider value={contextValue}>
-      {children}
+      {isLoadingOrFetching ? <></> : children}
     </JellyfinContext.Provider>
   );
 };
 ```

lostb1t:
dont have much experience with expo but we probaply want to wait with hiding the splash screen to after everything is loaded

https://github.com/streamyfin/streamyfin/blob/develop/app/_layout.tsx#L219

tom-heidenreich:
I fixed the issue with the Splash Screens now in a new PR #437. I'm closing this now, because this was just a quick and dirty solution

Issue #438, opened 2025-01-16
[Bug]: "Currently Streaming on" text cropped out


anubhavvs:
### What happened?

The text for "**Currently Streaming On**" appears to be cropped out due to overflow issues. This only happens for longer texts.

### Reproduction steps

1. Go to Search, then Discover
2. Search for a movie
3. Go to the Details section

### Which device and operating system are you using?

iPhone 16 with iOS 18.2

### Version

0.24.0

### If applicable, please add screenshots to help explain your problem. You can drag and drop images here or paste them directly into the comment box.

<img src="https://github.com/user-attachments/assets/cf12f90e-6dfa-47ab-b43b-fa7268613849" width="400">

Issue #392, opened 2025-01-07
[Bug]: Jellyseerr Intigration "Currently Streaming On" misaligned text


Panda260:
### What happened?

the text is not Alined when you search for any media with the new DIscovery page from the jellyseerr intigration

### Reproduction steps

1. Search Tab
2. Search a movie or TV Show it happend with "The Beekeeper"
3. then you see a misaligned text at Current Streaming On

### Which device and operating system are you using?

Xiaomi 14T Pro
Adndroid 14

### Version

0.24.0

herrrta:
<img width="422" alt="image" src="https://github.com/user-attachments/assets/5668f0ff-8cb2-401a-b38c-addbb9cd7351" />

herrrta:
fixed in referenced commit ^

anubhavvs:
@herrrta I made a PR with a better solution for this issue if you can have a look.

https://github.com/streamyfin/streamyfin/pull/439

herrrta:
This is closed since its been fixed.

If all you did was add flex grow then share the differences between current code & what you have in the beekeeper movie

anubhavvs:
Current solution:

<img src="https://github.com/user-attachments/assets/187c5448-539f-4048-a4d0-a576a9868e02" width="300" />

**The values are pushed down to the next row even if there is no overflow. This makes them not align together in the same height.**

My Solution:

<img src="https://github.com/user-attachments/assets/858d24ce-d900-445c-a4e0-d764480f536d" width="300" />

**Only the row with an overflowing value is wrapped to the next row instead of all the other fields. All the other fields are centered and aligned now.**

herrrta:
> @herrrta 
> 
> Current solution:
> 
> <img src="https://github.com/user-attachments/assets/187c5448-539f-4048-a4d0-a576a9868e02" width="300" />
> 
> **The values are pushed down to the next row even if there is no overflow. This makes them not align together in the same height.**
> 
> My Solution:
> 
> <img src="https://github.com/user-attachments/assets/858d24ce-d900-445c-a4e0-d764480f536d" width="300" />
> 
> **Only the row with an overflowing value is wrapped to the next row instead of all the other fields. All the other fields are centred and aligned now.**

Yeah made them all under the title to avoid what you just shared. All but the last are aligned now.

I wanted them all to be under the titl

anubhavvs:
> 
> Yeah made them all under the title to avoid what you just shared. All but the last are aligned now in your screenshot
[...]

PR #439, opened 2025-01-16
fix: text overflow for jellyseer

anubhavvs:
Fixed the issue where **"Currently Streaming on"** would overflow outside the visible boundaries of the screen.

#392 
#438

## Summary by Sourcery

Bug Fixes:
- Fixed text overflow for the "Currently Streaming On" section in Jellyseer.

### DIFF
--- a/components/jellyseerr/DetailFacts.tsx
+++ b/components/jellyseerr/DetailFacts.tsx
@@ -29,10 +29,10 @@ const Facts: React.FC<
 > = ({ title, facts, ...props }) =>
   facts &&
   facts?.length > 0 && (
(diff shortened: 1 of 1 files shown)


herrrta:
Add screenshots comparing latest in develop and what you have.

Seems like it would look the same?

herrrta:
Already been fixed so Title is misleading. There is no longer any overflow text.

You are proposing a different solution but your PR still keeps things misaligned. 
With your solutions, if other rows have long texts then things look even more misaligned. (staggered appearance)

Rather than current solution which is a consistent layout for any length of text

You should propose a different layout if what you care about is the very small loss of 'wasted' height

PR #443, opened 2025-01-17
Fix mark as played sheet logic being reversed

### LINKED ISSUE #435: Played/unplayed marking from context menu does not work
### What happened?

Played/unplayed marking from context menu does not work

### Reproduction steps

.

ios

### Version

0.25.0

_No response_

maxwellward:
fixes #435 

This issue was caused by the logic for marking content as played being backward. When you tapped "Mark not played" it would mark it as played, and vice versa.

I've changed the logic a little to make more sense and ensured it's updated in both the long-press sheet and the mark as played button at the top right when viewing a show.

## Summary by Sourcery

Bug Fixes:
- Correct the reversed behavior when marking content as played or not played through the long-press sheet and the button at the top right when viewing a show.

### DIFF
--- a/components/PlayedStatus.tsx
+++ b/components/PlayedStatus.tsx
@@ -46,7 +46,7 @@ export const PlayedStatus: React.FC<Props> = ({ item, ...props }) => {
       <RoundButton
         fillColor={item.UserData?.Played ? "primary" : undefined}
         icon={item.UserData?.Played ? "checkmark" : "checkmark"}
-        onPress={() => markAsPlayedStatus(!item.UserData?.Played || true)}
+        onPress={() => markAsPlayedStatus(!item.UserData?.Played)}
         size="large"
       />
     </View>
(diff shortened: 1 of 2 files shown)


lostb1t:
So, if this is being worked on. I wouldnt mind this having a state.

So only show "mark as played" if the media is unplayed and only show mark as unplayed when played

lostb1t:
When marking played with this pr I get the. following error "Invalid parameters for markAsPlayed"

lostb1t (changes requested):
Marking as played from a library view gives me an error

maxwellward:
Can you provide any further information such as a specific error message, screenshot, or video?

I can't seem to reproduce that issue (see attached video, I may be misreading your comment)

https://github.com/user-attachments/assets/ab5153c1-ce1b-44f0-8901-5346fe58ab80

PR #471, opened 2025-01-28
[android] fix: Revert change that broke self signed cert

RodoMa92:
Partially revert 252fc43 since it reverted a change needed to let the app accept user self signed certificates.

## Summary by Sourcery

Bug Fixes:
- Fixed an issue that prevented the app from accepting user-added certificates.


lostb1t:
has already been fixed

PR #533, opened 2025-02-15
feat: first version of AVAssetDownloadURLSession module for ios

fredrikburmester:


fredrikburmester:
I'm having some trouble trying to figure out the Android version of this.

Currently I'm using ExoPlayer to download the HLS stream and it's working so far. Right now the issue is getting the files in the correct directory and creating a m3u8 file that VLC can play.

GodTamIt:
> I'm having some trouble trying to figure out the Android version of this.
> 
> Currently I'm using ExoPlayer to download the HLS stream and it's working so far. Right now the issue is getting the files in the correct directory and creating a m3u8 file that VLC can play.


@fredrikburmester I just saw this.

Is there a reason the `DownloadHelper` + ExoPlayer path doesn't work here?

lostb1t:
this pr is not needed anymore because we have backgrojnd downloads working in latest beta

Issue #536, opened 2025-02-16
[Bug]: unable to log in to jellyseer with empty/no jellyfin password


velodromen:
### What happened?

Have a regular, non-root, account that has no password set.  I am able to login to jellyfin with this account. However when i try to set up the jellyseerr connection I receive the error: "Failed to login". 

I really would like to avoid setting a password since the family uses multiple profiles on LG TVs and its quite tedious to login with PW using a remote every time you want your own profile.. 

### Reproduction steps

In the Android app I go to "Settings" and then "Jellyseerr" under plugins.
I then fill in my Jellyseerr URL (have tried with both DNS name and IP adress, no difference) and hit save. I then press login underneath and receive the error message.

The logs state invalid credentials. When logging in to Jellyseerr via WebUI using the same username and no password it works.

### Which device and operating system are you using?

Android phone, latest Andoid OS.

### Version

0.26.1

sbaiahmed1:
I will create a pull today to fix it! already found a fix.
Stay tuned

PR #540, opened 2025-02-17
fix: handle empty password in Jellyseerr login

sbaiahmed1:
handle empty password in Jellyseerr login and update state initialisation.
Targeted Passwordless accounts & enabled their login

I did prefer `useState<string>("")` over `useState<string | undefined>(undefined)` because:
1. Avoids `undefined` complexity; empty string (`""`) is simpler for checks and form handling.
2. Ensures type consistency (`string` instead of `string | undefined`).
3. Aligns with APIs/form libraries that use `""` for empty values.
4. Reduces bugs by avoiding `undefined` edge cases.
we use `undefined` only if explicitly distinguishing "no value" from "empty value" is necessary.

Resolves issue : https://github.com/streamyfin/streamyfin/issues/536

DEMO: 

https://github.com/user-attachments/assets/4611e00c-cef2-43ff-a673-8bc7c5485298

## Summary by Sourcery

Allow users with empty passwords to log in to Jellyseerr. Update the password state initialization to handle empty strings.

Bug Fixes:
- Fix login issues caused by empty passwords in Jellyseerr, enabling passwordless accounts to log in successfully.
- Correct the state initialization for the Jellyseerr password field to allow empty strings, addressing issues with passwordless logins and improving overall login handling

PR #547, opened 2025-02-19
fix: improve Jellyseerr login handling and state management

sbaiahmed1:
handle empty password in Jellyseerr login and update state initialisation.
Targeted Passwordless accounts & enabled their login

I did prefer `useState<string>("")` over `useState<string | undefined>(undefined)` because:
1. Avoids `undefined` complexity; empty string (`""`) is simpler for checks and form handling.
2. Ensures type consistency (`string` instead of `string | undefined`).
3. Aligns with APIs/form libraries that use `""` for empty values.
4. Reduces bugs by avoiding `undefined` edge cases.
we use `undefined` only if explicitly distinguishing "no value" from "empty value" is necessary.

I did also remove the **settings.jellyseerrServerUrl** in the mutation condition because it is useless, we are getting the **settings.jellyseerrServerUrl** already and putting it in the **jellyseerrServerUrl** state value

Resolves issue : https://github.com/streamyfin/streamyfin/issues/536

continuation for PR: https://github.com/streamyfin/streamyfin/pull/540

DEMO: 

https://github.com/user-attachments/assets/4611e00c-cef2-43ff-a673-8bc7c5485298

### DIFF
--- a/components/settings/Jellyseerr.tsx
+++ b/components/settings/Jellyseerr.tsx
@@ -12,6 +12,7 @@ import { Input } from "../common/Input";
 import { Text } from "../common/Text";
 import { ListGroup } from "../list/ListGroup";
 import { ListItem } from "../list/ListItem";
+import {URL_REGEX} from "@/constants/Regex";
 
 export const JellyseerrSettings = () => {
   const {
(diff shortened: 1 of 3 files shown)


herrrta:
Not relavant since we treat the value as a truthy value. blank or undefined the "complexity" you state doesnt change

herrrta:
we already allow for blank password logins too

Issue #595, opened 2025-03-10
Ability to consume webhook notifications and forward to clients


herrrta:
**Describe the solution you'd like**
I want to be able to consume webhooks from software like:

- jellyseerr
- radarr
- sonarr
etc

and forward any of these to my users

[see required plugin issue](https://github.com/streamyfin/jellyfin-plugin-streamyfin/issues/26)

PR #596, opened 2025-03-10
feat: Ability to consume webhook notifications and forward to clients…

herrrta:
… #595

- foward expo device tokens to users plugin instance

## Summary by Sourcery

Enhancements:
- Forwards expo device tokens to the users plugin instance.


herrrta:
I still have to verify a few things only you can see @fredrikburmester in expo dashboard

PR #632, opened 2025-03-31
fix: conditionally hide the jellyseer report issue button

### LINKED ISSUE #631: [Bug]: Jellyseer report issue button always shown
### What happened?

For users that do not have permission to request items on jellyseer, the report issue button is shown on all jellyseer items, including ones that haven't even been requested yet/don't exist on jellyfin. The report button should not be visible if the item has not yet been request/doesn't exist on jellyfin. This bug occurs because if a user does not have request permissions, the report issue button is shown instead of the request button.

A) This just makes logical sense and showing a report issue button that doesn't do anything is misleading to the end user.
B) This matches the functionality of the jellyseer website.

### Reproduction steps

1. log in to streamyfin and then the jellyseer integration as a user that does NOT have request permissions 
2. search for an item that you already have on jellyfin - the report issue button will be there which is good/correct
3. search for an item you are certain you do not already have on jellyfin - the report button will be there which is the bug in question

found while doing related development work on the emulated device on my laptop.

### Version

0.28.0

_No response_

a-dubs:
Fixes #631

## Summary by Sourcery

Conditionally hide the Jellyseerr report issue button based on media information availability

Bug Fixes:
- Modify the visibility of the report issue button to only show when a Jellyfin media ID is present

Enhancements:
- Add consistent margin to buttons for improved layout spacing

### DIFF
--- a/app/(auth)/(tabs)/(home,libraries,search,favorites)/jellyseerr/page.tsx
+++ b/app/(auth)/(tabs)/(home,libraries,search,favorites)/jellyseerr/page.tsx
@@ -236,18 +236,18 @@ const Page: React.FC = () => {
                   }}
                 />
               </View>
-              <View className='mb-4'>
+              <View>
                 <GenreTags genres={details?.genres?.map((g) => g.name) || []} />
               </View>
               {isLoading || isFetching ? (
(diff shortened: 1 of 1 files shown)


a-dubs:
Closing in favor of #634

PR #636, opened 2025-04-04
feat: adding reportPlaybackStart which allows tracking to work well

lancechant:
Adding this seems to allow the tracking to be picked up and work correctly

## Summary by Sourcery

Improve playback tracking functionality in the direct player component by adding proper reporting of playback start and handling stream readiness

New Features:
- Add a new state variable to track stream readiness before reporting playback start

Bug Fixes:
- Ensure playback tracking is correctly initiated when stream is fully loaded

Enhancements:
- Implement reporting of playback start when stream becomes ready
- Add more detailed logging and tracking for playback state changes

### DIFF
--- a/app/(auth)/player/direct-player.tsx
+++ b/app/(auth)/player/direct-player.tsx
@@ -24,6 +24,7 @@ import {
   type MediaSourceInfo,
   PlaybackOrder,
   type PlaybackProgressInfo,
+  PlaybackStartInfo,
   RepeatMode,
 } from "@jellyfin/sdk/lib/generated-client";
 import {
@@ -151,6 +152,7 @@ export default function page() {
   }
 
   const [stream, setStream] = useState<Stream | null>(null);
+  const [streamReady, setStreamReady] = useState(false);
   const [streamStatus, setStreamStatus] = useState({
     isLoading: true,
     isError: false,
@@ -191,6 +193,7 @@ export default function page() {
           result = { mediaSource, sessionId, url };
         }
         setStream(result);
+        setStreamReady(true);
       } catch (error) {
         console.error("Failed to fetch stream:", error);
         setStreamStatus({ isLoading: false, isError: true });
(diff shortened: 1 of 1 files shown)


Simon-Eklundh:
I've tested this, it fixes the scrobbliing to Trakt, as well as the media cleanup plugin deletion I've been missing :D

haven't tested extensively but it at least does those two things which is great :D

app/(auth)/player/direct-player.tsx
```diff
@@ -201,13 +204,34 @@ export default function page() {
...
+      await getPlaystateApi(api!).reportPlaybackStart({
+        playbackStartInfo: currentPlayStateInfo() as PlaybackStartInfo,
+      });
+    };
+
+    reportPlaybackStart();
+  }, [streamReady]);
+
   const togglePlay = async () => {
+    console.log("I'm triggered");
```
Simon-Eklundh, line 220: might wanna remove this log, I assume was for debugging

app/(auth)/player/direct-player.tsx
```diff
@@ -201,13 +204,34 @@ export default function page() {
...
     lightHapticFeedback();
     setIsPlaying(!isPlaying);
+    writeToLog("ERROR", `${isPlaying}`, isPlaying);
```
Simon-Eklundh, line 223: possibly this too?

app/(auth)/player/direct-player.tsx
lancechant: Lol I thought I removed it, will update now

app/(auth)/player/direct-player.tsx
```diff
@@ -151,6 +152,7 @@ export default function page() {
   }
 
   const [stream, setStream] = useState<Stream | null>(null);
+  const [streamReady, setStreamReady] = useState(false);
```
lostb1t, line 155: Why not use reuse stream prop? if not null then its started.

lancechant: Was just trying to separate logic, and not affect other items, Will change and use the stream😁

PR #660, opened 2025-04-23
fix: dolby vision on supported devices, specifically profile 5

lancechant:
Fixing dolby vision profile 5 (since it doesn't have a fallback) to only work on supported devices
Removed some unused files

## Summary by Sourcery

Implement device-specific Dolby Vision support by dynamically generating device profiles based on device capabilities

Bug Fixes:
- Fix Dolby Vision profile 5 handling by restricting playback on unsupported devices

Enhancements:
- Add dynamic device profile generation with platform-specific Dolby Vision detection
- Improve codec profile configuration with conditional support

Chores:
- Remove unused device profile files for Android and iOS


lancechant:
@fredrikburmester I've removed some unused files, but not sure if they were to be used in the future, if you can let me know so I can add them back?

lostb1t:
@retardgerman any change you might be able to test this?

retardgerman:
@lostb1t how should I do that? Simulator or my device?

lostb1t:
@retardgerman device if possible

retardgerman:
All the basic features work. Unfortunately, I don't have any DoVi Profile 5 material available.

lostb1t:
@retardgerman thanks for checking m8!

retardgerman:
@lostb1t I'm doing my Part!

nyanmisaka:
@lancechant VLC doesn't actually support Dolby Vision Profile 5 yet. So unless you're using Apple's native player, you should mark DOVI as unsupported on all devices.

lancechant:
Hey @nyanmisaka thanks, I was aware of that when I was doing some research on it, of course aren't an expert on video quality so did a best bet assumption. The only reason I didn't block it outright was

1. Incase it is supported in the future then it should just work (hopefully)
2. The video player does change with streamyfin again.

I looked around and also tried to match as best as I could assume on how jellyfin does it with it's native clients just to have some parity, because support for vision and audio are on a device level too not just a player from what I understood.

nyanmisaka:
> Hey @nyanmisaka thanks, I was aware of that when I was doing some research on it, of course aren't an expert on video quality so did a best bet assumption. The only reason I didn't block it outright was
> 
> 1. Incase it is supported in the future then it should just work (hopefully)
> 2. The video player does change with streamyfin again.
> 
> I looked around and also tried to match as best as I could assume on how jellyfin does it with it's native clients just to have some parity, because support for vision and audio are on a device level too not just a player from what I understood.

I wish so, but the fact is that Dolby and open source software just don't go together.

A few examples: Microsoft Edge on Windows requires the paid Dolby Vision plugin and some experimental options to make Profile 5 work. Profile 5 in the libplacebo renderer used by MPV or VLC was done through clean room reverse engineering, and it requires a lot of device GPU performance, which means heat and low battery life on mobile devices. As for Profile 5 in Apple's decoder and renderer, there is no public API to use it.

So IMHO it's better to switch to reliable server transcoding and solve the strange colors brought by Profile 5 once and for all. If Dolby really changes its mind and wants to open up this proprietary format in the future, it's not too late to enable it selectively on supported devices.

PR #665, opened 2025-04-28
fix: improve empty state layout in library view

Ryan0204:
Fixed header filter disappearing.

From Discord: 
> Question about the library/collection view.. using the genre and year filter works great. But when using a filter combination, setting both genre and year, that results in no items found the whole filter menu disappears.. rendering the collection view unusable? Do more people notice the same?

### DIFF
--- a/app/(auth)/(tabs)/(libraries)/[libraryId].tsx
+++ b/app/(auth)/(tabs)/(libraries)/[libraryId].tsx
@@ -433,14 +433,18 @@ const Page = () => {
       </View>
     );
 
-  if (flatData.length === 0)
+  if (flatData.length === 0) {
     return (
-      <View className='h-full w-full flex justify-center items-center'>
-        <Text className='text-lg text-neutral-500'>
-          {t("library.no_items_found")}
-        </Text>
+      <View className='h-full w-full mt-24'>
+        {ListHeaderComponent()}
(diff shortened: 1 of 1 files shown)


fredrikburmester:
Awesome! Is it ready for merge?

Ryan0204:
Yes! However, there should have a better way to handle the spacing between top of the screen and header. 

Currently I am using `mt-24`, which looks good while I'm testing on `iPhone 16 Pro`.

app/(auth)/(tabs)/(libraries)/[libraryId].tsx
```diff
@@ -433,14 +433,18 @@ const Page = () => {
       </View>
     );
 
-  if (flatData.length === 0)
+  if (flatData.length === 0) {
```
Comment (lostb1t) [inline app/(auth)/(tabs)/(libraries)/[libraryId].tsx:436]: Im gonna suggest removing this whole block.

lostb1t (changes requested):

app/(auth)/(tabs)/(libraries)/[libraryId].tsx
Comment (Ryan0204) [reply app/(auth)/(tabs)/(libraries)/[libraryId].tsx]: By default <FlashList> will show no result already, so yes I have removed it. Thanks for the suggestion! 

ScreenShot-2025-04-29 at 00 05 06@2x

PR #675, opened 2025-05-03
feat: move to custom download handler with background download support

lostb1t:
Replace ffmpeg with a transcoded file stream using background tasks

## Summary by Sourcery

Replace FFmpeg-based download handling with a custom download method using background tasks and optimized server streaming

New Features:
- Add support for optimized server download method

Bug Fixes:
- Improve download handling for different server configurations

Enhancements:
- Modify download process to use background tasks
- Improve stream URL handling for downloads

Chores:
- Remove FFmpeg-related dependencies

## Summary by Sourcery

Replace FFmpeg-based download handling with a custom download method using background tasks and optimized server streaming

New Features:
- Add support for optimized server download method

Bug Fixes:
- Improve download handling for different server configurations

Enhancements:
- Modify download process to use background tasks
- Improve stream URL handling for downloads

Chores:
- Remove FFmpeg-related dependencies


utils/jellyfin/media/getStreamUrl.ts
```diff
@@ -70,9 +72,12 @@ export const getStreamUrl = async ({
 
   sessionId = res.data.PlaySessionId || null;
   mediaSource = res.data.MediaSources[0];
-  const transcodeUrl = mediaSource.TranscodingUrl;
+  let transcodeUrl = mediaSource.TranscodingUrl;
 
   if (transcodeUrl) {
+    if (forceStream) {
+      transcodeUrl = transcodeUrl.replace("master.m3u8", "stream");
```
lancechant, line 79: @lostb1t since the transcode URL is changing here, I think the directPlay file is checking for that 3u8 string as well to the playback reporting, so that might have to change as well? Or not?

lostb1t: direct play already is a file stream not hls. Tho it doesnt hurt to apply it anyway i guess. It might not be in edge cases

lancechant: I was talking about the player itself, on this line to be exact: [direct-player.tsx](https://github.com/streamyfin/streamyfin/blob/f4750e781d2894afdf1eabccc37e3eeec9d8cbaa/app/(auth)/player/direct-player.tsx#L267) as it uses that `m3u8` to report if it's streaming or not

lostb1t: yeah but by default forceStream is false so the player gets the m3u8 which is the current default. Not sure theres a reason to change that now.

PR #696, opened 2025-05-16
fix: Fixed container name mp4 in transcoding profiles

### LINKED ISSUE #585: [Bug]: Invalid HLS segment container
### What happened?

Intermittently, when trying to begin a transcoded stream from Jellyfin, the Streamyfin player will just sit on 0:00 and not play. Checking the Jellyfin logs, this line will be found:

`[ERR] [68] Jellyfin.Api.Controllers.DynamicHlsController: Invalid HLS segment container: "fmp4", default to mpegts`

When this occurs, the stream will not start. However, once you exit the player and press play again, it will (often) start correctly the second time.

### Reproduction steps

1. Select a film from your Jellyfin library.
2. Select a transcoding quality level.
3. Press play.
4. Intermittently (can't figure out a way to reliably trigger it), the player will stay stuck on 0:00 and the video will not play (stays a black screen).
5. Check your Jellyfin logs. Look for the line `[ERR] [68] Jellyfin.Api.Controllers.DynamicHlsController: Invalid HLS segment container: "fmp4", default to mpegts`

Server:
- Unraid 7.0.1
- Jellyfin 10.10.6 (jellyfin/jellyfin repo)

Client:
- iPhone 13 Pro Max
- iOS 18.3.1

### Version

0.26.1

_No response_

nyanmisaka:
The HLS container should either be `mp4` or `ts`.

See also: jellyfin-web client https://github.com/jellyfin/jellyfin-web/blob/5d9cda6dc917f04ba9b11f71a3137cdcc8dfbdab/src/scripts/browserDeviceProfile.js#L829

Should fixes: https://github.com/streamyfin/streamyfin/issues/585

## Summary by Sourcery

Bug Fixes:
- Correct the HLS container setting from "fmp4" to "mp4" in the native device profile

### DIFF
--- a/utils/profiles/native.js
+++ b/utils/profiles/native.js
@@ -79,7 +79,7 @@ export const generateDeviceProfile = async () => {
         Type: MediaTypes.Video,
         Context: "Streaming",
         Protocol: "hls",
-        Container: "fmp4",
+        Container: "mp4",
         VideoCodec: "h264, hevc",
         AudioCodec: "aac,mp3,ac3,dts",
       },


lancechant:
@nyanmisaka just a question is fmp4 actually a thing? as I wanted to change this as well because of server logs saying it doesn't exist, but still worked, so any insight would be great, just from a learning point😅

nyanmisaka:
> @nyanmisaka just a question is fmp4 actually a thing? as I wanted to change this as well because of server logs saying it doesn't exist, but still worked, so any insight would be great, just from a learning point😅

`fmp4` is a FFmpeg [option](https://github.com/jellyfin/jellyfin-ffmpeg/blob/0dd5381dc544e6c83735afa1434a9b1e536d1ced/libavformat/hlsenc.c#L3165), not a container suffix. Jellyfin just falls back to the legacy `mpegts/.ts` when the client enters an incorrect value.

According to Apple's HLS [documentation](https://developer.apple.com/documentation/http-live-streaming/hls-authoring-specification-for-apple-devices#Delivery), the container suffix for fMP4 segments should be `.mp4` or `.m4s`. Jellyfin uses `.mp4`, and when the client requests `.mp4`, the transcoder will set the `hls_segment_type` option to `fmp4`.

lancechant:
Awesome thanks for the clarification :)

PR #703, opened 2025-05-20
fix: always transcode with profile 5

lancechant:
changed logic to always get the server to transcode with profile 5 content from dicsussion on #660

## Summary by Sourcery

Always exclude Dolby Vision content by unconditionally adding the ‘NotEquals VideoRangeType=DOVI’ condition to the native video profile and removing all device capability checks

Bug Fixes:
- Enforce use of profile 5 for all transcoding by always excluding Dolby Vision

Enhancements:
- Remove supportsDolbyVision helper and related conditional branching

## Summary by CodeRabbit

- **Refactor**
  - Updated device profile logic to always exclude Dolby Vision content, removing dynamic checks for Dolby Vision support.
- **Chores**
  - Simplified internal logic for determining supported video codecs.

### DIFF
--- a/utils/profiles/native.js
+++ b/utils/profiles/native.js
@@ -7,28 +7,7 @@ import DeviceInfo from "react-native-device-info";
  */
 import MediaTypes from "../../constants/MediaTypes";
 
-// Helper function to detect Dolby Vision support
-const supportsDolbyVision = async () => {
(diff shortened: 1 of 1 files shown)


lancechant:
@nyanmisaka here is the change as discussed

PR #721, opened 2025-06-02
fix: made tv os compile

lancechant:
Made useDownload return no-op functions instead of throwing an error when running tvos

## Summary by Sourcery

Bug Fixes:
- Return no-op download functions on TV platforms instead of throwing an exception to support tvOS

### DIFF
--- a/providers/DownloadProvider.tsx
+++ b/providers/DownloadProvider.tsx
@@ -842,12 +842,39 @@ export function DownloadProvider({ children }: { children: React.ReactNode }) {
 }
 
 export function useDownload() {
+  if (Platform.isTV) {
+    // Since tv doesn't do downloads, just return no-op functions for everything
+    return {
+      processes: [],
+      startBackgroundDownload: useCallback(
+        async (
+          _url: string,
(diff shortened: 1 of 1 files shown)


lancechant:
So just an if here for platformTV not sure on the code block there, but would that be okay? [line-461](https://github.com/streamyfin/streamyfin/blob/1b812ebed5a6961680c62791199624ed68d88b6e/app/_layout.tsx#L461)

The `useDownload` is used in a lot more places so that's why I used no-ops instead of conditional in the other locations

lostb1t:
i understand . But just making it not work will bite us in the ass because we will have to remove that at one point anyway. 

A button that uses it wil now just not work. Better to be abead if the game?

lancechant:
I'm happy to do the following:

```
let value = null
if (platform.isTV {
value = useDownload.value
}
```
But that was across about 10 different files as the `useDownload` function was called directly and not `useDownloadProvider` ?

Am I missing the mark maybe? my brain could be fried from the long day?

lostb1t:
merged. we will make it better when the time comes

Issue #824, opened 2025-06-30
[Bug]: Error at playing videos


arch-fan:
### What happened?

The following issue is happening:
<img width="394" alt="Image" src="https://github.com/user-attachments/assets/2ae80e29-0044-44e0-92a1-af1ba2f7d8b5" />

This says that an error happened playing the video. This is happening to me at a downloaded video.

### Reproduction steps

1. Download a video
2. Play it

### Which device and operating system are you using?

iPhone 14, iOS 18.5

### Version

0.28.0

Simon-Eklundh:
This is a known error

PR #825, opened 2025-06-30
fix: error at playing due to unload video

arch-fan:
Fixes the described [issue](https://github.com/streamyfin/streamyfin/issues/824)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved video player behavior by ensuring it only appears when a valid stream is available.


Simon-Eklundh:
I believe this is already solved in the latest beta (both in develop and in main)

(in a different area of the code)

PR #846, opened 2025-07-11
Update package json from expo doctor update

Alexk2309:
## Summary by CodeRabbit

* **Chores**
  * Updated dependency version for improved compatibility.
  * Modified linting configuration to prevent errors when no files match certain patterns.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "@react-navigation/bottom-tabs": "^7.2.0",
     "@react-navigation/material-top-tabs": "^7.1.0",
     "@react-navigation/native": "^7.0.14",
-    "@shopify/flash-list": "1.8.3",
+    "@shopify/flash-list": "1.7.3",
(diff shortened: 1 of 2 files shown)


Alexk2309:
Expo doctor was having problems this seems to the fix it suggested

package.json
```diff
@@ -138,7 +138,7 @@
   },
   "lint-staged": {
     "*.{js,jsx,ts,tsx}": [
-      "biome check --write --unsafe"
+      "biome check --write --unsafe --no-errors-on-unmatched"
```
Alexk2309, line 141: I think this is needed. When biome doesn't find any to run against. I don't think we should consider this an error.

lancechant: I didn't delve into the error too much, but it looked like it wasn't linting on the application code because of the updated config with the upgrade, but I could be wrong it's been a few weeks now, if it's linting then I guess it's working..

PR #852, opened 2025-07-12
Revert "fix: Recently Added isn't updating correctly."

Alexk2309:
Reverts streamyfin/streamyfin#686

I'm actually encountering bugs with this implementation. It feels a bit hacky  while it might work in some scenarios, it's not a clean or reliable solution. I’m not a fan of introducing a workaround like this, especially when it adds unnecessary code. We should aim to stick to the SDK and rely on the library’s official API, even if it's not perfect.

## Summary by CodeRabbit

* **New Features**
  * Simplified the sorting options for libraries by removing conditional filtering based on collection type.
  * Updated the latest media section on the home screen to use a more streamlined data source and improved parameter handling for custom sections.

* **Bug Fixes**
  * Removed the "Date Episode Added" sort option from all library views.

* **Style**
  * Removed the watched indicator from series posters for a cleaner display.


Alexk2309:
IMG_8895

IMG_8896

Issue #858, opened 2025-07-16
[Bug]: Casting restarts episode from beginning instead of resuming from last watched position


RomainBonneau:
### What happened?

When I try to cast an episode from Streamyfin, it always restarts the episode from the beginning instead of resuming from where I previously stopped watching. This issue occurs consistently and makes it difficult to continue watching content smoothly across devices.

Expected behavior:
The episode should resume playback from the last watched position when casting.

### Reproduction steps

- Watch part of an episode on Streamyfin and stop playback (e.g., halfway through).

- Later, try to cast the same episode to a Chromecast or other casting device.

- The episode starts from the beginning instead of resuming from the last watched timestamp.

### Which device and operating system are you using?

Platform: iOS

Casting device: Chromecast

### Version

0.28.0

n0vella:
This is still a problem for me with latest play store version

HenrysCat:
Still not working on IOS either.

PR #868, opened 2025-07-20
fix: convert playback position ticks to seconds in PlayButton component

JasBogans:
Reference to this issue: https://github.com/streamyfin/streamyfin/issues/858

## Summary by CodeRabbit

* **New Features**
  * Media playback now resumes from your last watched position when casting or using supported playback sessions; if no previous position is available, playback starts from the beginning.
  * Resume position is calculated automatically so playback continues seamlessly without changing stream behavior.
  * Improves the resuming experience across sessions while preserving existing playback and streaming behavior.


Gauvino:
fix linting issue and it's all good !

Gauvino (dismissed):

Gauvino (approved):

Gauvino:
@JasBogans tested this pr and still it doesn't fix the problem, it seem like the playback is not tracked

fredrikburmester:
> @JasBogans tested this pr and still it doesn't fix the problem, it seem like the playback is not tracked

Playback can't be tracked with chromecast. But this PR only tries to fix the **starting** progress, no? If so then we should only test that.

Gauvino:
Exactly but still it doesn't work

fredrikburmester:
Hmm ok, i'll have to test it also.

Gauvino:
> Hmm ok, i'll have to test it also.

Test this ipa :

https://github.com/streamyfin/streamyfin/actions/runs/17055928661/artifacts/3794032798

PR #912, opened 2025-08-08
fix: intro skipper

lancechant:
Fixed an issue where the seek was not initilized when viewing Added the same checks to the rest of the functions

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of player controls with safer, logged handling for play, pause, seek, picture‑in‑picture, and audio/subtitle selection.
  * More robust volume adjustments with preserved error logging.

* **Refactor**
  * Centralized and consistent error handling for all player interactions.

* **Performance**
  * Debounced progress reporting and reduced frequency of playback-position URL updates (now updated every ~5s) to lower network activity and improve responsiveness.


Simon-Eklundh:
this is a test comment

edit: this was me trying to figure out how approves work on github :D

Simon-Eklundh:
I tested this, while it does add the skipIntro  functionality back in, it also seemed to make my testing device (old android phone) lag a lot more than what it does with the current beta, so I'd be hesitant to approve as is

lancechant:
Yeah I did notice that, will have to see what can be improved, as that was pretty much existing functionality?

Simon-Eklundh:
Yeah, it's interesting, off the top of my head I have no clue how that made a difference.... (Mind you, I didn't read through your pr properly)

lancechant:
Hey @Simon-Eklundh would you see if there is any improvements now?

Simon-Eklundh:
Tested it (albeit quite quickly) and the lag seems gone :D good job!

Will give it a proper code review later tho

Alexk2309:
I don’t think this is needed anymore. The change adds a lot of code, which isn’t ideal, and it feels like a band-aid that masks the real issue. Why does seeking fail only during “Skip Intro” when it works everywhere else? We should dig deeper and address the root cause instead of patching around it. I think there probably underlying reason with our design that causes to need to do this.

PR #968, opened 2025-08-21
feat: replace content item dropdowns with sheets

lostb1t:
- Replaced drop-downs on item content page with bottom sheets. Reusing the existing filtersheet.
- Search now only automatically enabled if more then 15 items.

Did not replace drop downs on other places like download sheet. Should we?

## Summary by CodeRabbit

- **New Features**
  - UI now uses sheet-style pickers for quality, video source, audio and subtitle tracks for a consistent selection experience.

- **Bug Fixes**
  - Audio submenu is omitted when no audio tracks exist to reduce clutter.

- **Documentation**
  - Filter search now auto-displays for long lists and can be disabled; docs updated to reflect the new behavior.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,4 @@ streamyfin-4fec1-firebase-adminsdk.json
 .env.local
 *.aab
 /version-backup-*
+bun.lockb
--- a/components/BitRateSheet.tsx
+++ b/components/BitRateSheet.tsx
@@ -1,0 +1,122 @@
+import { useMemo, useState } from "react";
+import { useTranslation } from "react-i18next";
+import { Platform, TouchableOpacity, View } from "react-native";
+import { Text } from "./common/Text";
+import { FilterSheet } from "./filters/FilterSheet";
+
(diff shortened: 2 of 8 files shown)


fredrikburmester:
When opening the bitrate selector first time set to "max" it is not selected in the bottom sheet list.
Simulator Screenshot - iPhone 16 Pro - 2025-08-21 at 14 14 06

fredrikburmester:
The search is missing a border: 
Screenshot 2025-08-21 at 14 14 54

.gitignore
```diff
@@ -46,3 +46,4 @@ streamyfin-4fec1-firebase-adminsdk.json
 .env.local
 *.aab
 /version-backup-*
+bun.lockb
```
Gauvino, line 49: Why adding bun.lockb since we don't have one we only have bun.lock right ?

lostb1t: for some reason it always generates a bun.lockb for me

Gauvino: Wtffff

PR #972, opened 2025-08-22
fix: tv login layout

lancechant:
Fixed the tv login layout to be a bit easier to navigate and use

image

## Summary by CodeRabbit

- New Features
  - Added a dedicated TV experience for Login with platform-aware layout, server connect flow, server discovery and previously used servers, plus Quick Connect.
  - TV buttons now support focus states with smooth scale animations and visual glow.
  - Inputs gain TV-focused styling with clearer focus/blur states and improved placeholder visibility.
- Style
  - Refined purple button appearance with focused/unfocused variations.
  - Improved contrast and responsiveness for inputs on TV.
- Notes
  - Mobile experience remains unchanged.


kamilkosek:
Hi @lancechant , does it work with dpad navigation only ? Dont know if it is in my branch only, but actually I cant login with only the remote.

lancechant:
Hey @kamilkosek it does work with the remote navigation yes, ensured it was all working before I merged it.

kamilkosek:
Thanks for the very quick response. In such case I have messed up something 😥

lancechant:
No problem at all, but in my experience it's always a package that causes it to break TV navigation, so I'd check what packages you might have added or that were upgraded from this pr. If you still don't come right, I'll take a look tomorrow as well 😁

kamilkosek:
I suspect the `TVDrawerLayout`  ... will investigate

PR #1032, opened 2025-09-03
fix: Jellyseerr discovery crash

### LINKED ISSUE #1024: [Bug]: Application crash when opening "Discover" section

### What happened?

When opening Streamyfin's search page's discover section, the app crashes with no error being displayed in the logs.

The app should open the search page without crashing.

### Reproduction steps

1. Open the app's search Discovery section
2. Crash

LineageOS 20

### Streamyfin Version

0.30.2

Jellyfin and Jellyseer running on a Docker container with the latest version on a Debian server

### Screenshots or Videos

_No response_

```shell
Nothing in the app
```

### Additional information

After toying around with ProxyPin to analyse the requests being made on my device, it seems that the app is being given a movie recommendation with no poster, tries to make a request to `https://server_url/images/overseerr_poster_not_found_logo_top.png` which redirects to `/login` for some reason.
When opening the URL in my browser, it returns a 404 error. Replacing the path with `/images/overseerr_poster_not_found_logo_top.png`

stenlan:
# 📦 Pull Request

## 🔖 Summary
Fix the client crashing when opening the Jellyseerr discovery page.

## 🏷️ Ticket / Issue
Fixes #1024 

## 🛠️ What’s Changed
I have removed some typecasting that was hiding these type errors, although I am still not fully happy with this fix. We now filter most items passed to the `JellySeerrPoster` component. It can also handle an extra type: `PersonCreditCast`. I've also added extra null coalescing in case we still receive unexpected data from the server.

I personally think the MediaType enum in jellyseerr/server/constants/media.ts should be updated to add the Person and Collection type as well. Lastly, for some reason my editor isn't picking up the `mediaInfo` property type in case `isJellyseerrMovieOrTvResult` is false:

image

even though it correctly understands that at this point, `item` is only `TvDetails | MovieDetails | PersonCreditCast | undefined`:

image

all of which have the `mediaInfo` property of type `Media` (defined in jellyseerr/server/entity/Media.ts). I don't know what causes this, maybe it has to do with the fact that it's a submodule, since my editor is giving some errors on its tsconfig.json.

Anyway, the issue should be fixed.

I've also cleaned up the useSettings hook to return an object instead of an array, so it's much cleaner, less error prone, and you don't need to retrieve unneeded properties if you need ones after them.

## Summary by CodeRabbit

- New Features
  - Browsing and poster views now support person-credit items (people appear in lists/posters).

- Bug Fixes
  - Company, genre and discovery lists now show only movie/TV entries.
  - Title, year and media-type display are more consistent across items.
  - Reduced rendering quirks by simplifying item handling.

- Refactor
  - Unified filtering and item routing for consistent browsing behavior.

### DIFF
--- a/app/(auth)/(tabs)/(home)/settings/hide-libraries/page.tsx
+++ b/app/(auth)/(tabs)/(home)/settings/hide-libraries/page.tsx
@@ -12,7 +12,7 @@ import { apiAtom, userAtom } from "@/providers/JellyfinProvider";
 import { useSettings } from "@/utils/atoms/settings";
 
 export default function page() {
-  const [settings, updateSettings, pluginSettings] = useSettings(null);
+  const { settings, updateSettings, pluginSettings } = useSettings();
   const user = useAtomValue(userAtom);
   const api = useAtomValue(apiAtom);
 
--- a/app/(auth)/(tabs)/(home)/settings/jellyseerr/page.tsx
+++ b/app/(auth)/(tabs)/(home)/settings/jellyseerr/page.tsx
@@ -3,7 +3,7 @@ import { JellyseerrSettings } from "@/components/settings/Jellyseerr";
 import { useSettings } from "@/utils/atoms/settings";
 
 export default function page() {
-  const [_settings, _updateSettings, pluginSettings] = useSettings(null);
+  const { pluginSettings } = useSettings();
 
   return (
     <DisabledSetting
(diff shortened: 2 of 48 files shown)


fredrikburmester:
We currently have a Jellyseerr authentication/cookies or something issue on develop. Have you experienced that, or fixed that with this PR also? Example, the "discover" button is missing even though you log into Jellyseerr in the settings.

stenlan:
> We currently have a Jellyseerr authentication/cookies or something issue on develop. Have you experienced that, or fixed that with this PR also? Example, the "discover" button is missing even though you log into Jellyseerr in the settings.

Yeah I experienced the same, I made these changes on the latest release branch, then rebased off the develop branch. I figure since they're separate bugs, and this is actually fully a type safety bug that can be statically analysed (after removing the typecasts), that they can stay separated.

PR #1036, opened 2025-09-04
fix: section titles

lostb1t:
## Summary by CodeRabbit

- Bug Fixes
  - Custom Home sections now use the section’s own title for identification, ensuring titles and translations display correctly.
  - Resolves issues where sections could appear with mismatched labels or duplicate identifiers.
  - Improves consistency of section naming across languages without changing underlying content or data behavior.


Simon-Eklundh (approved):
Can't break things more than they already are at least ;)

jk

PR #1051, opened 2025-09-09
fix: add better offline messages and checks

lancechant:
# 📦 Pull Request

## 🔖 Summary

- Changed logic to check for connection to the server than the internet Added better messages to show the difference
updated react-native-reanimated

## 🛠️ What’s Changed

- Type: fix
- Short summary: Added better checks and messages for different connection states
-->

## ✅ Checklist

- [ ] Code follows project style and passes lint/format (`npm|pnpm|yarn|bun` scripts)
- [ ] Type checks pass (tsc/biome/etc.)
- [ ] No secrets/credentials included; env vars documented
- [ ] Verified locally that changes behave as expected

## Summary by CodeRabbit

- New Features
  - Home screen now distinguishes no-internet, checking-server, and server-unreachable states with localized titles/subtitles.
  - Retry button localized and shows a refresh icon when available; improved loading feedback.
  - Background server reachability checks to more reliably decide when to show main content vs offline states.
- Bug Fixes
  - Adjusted authenticated navigation logic to avoid unintended redirects.
- Chores
  - Updated dependency: react-native-reanimated → ~3.19.1.

### DIFF
--- a/components/settings/HomeIndex.tsx
+++ b/components/settings/HomeIndex.tsx
@@ -74,7 +74,12 @@ export const HomeIndex = () => {
 
   const { getDownloadedItems, cleanCacheDirectory } = useDownload();
   const prevIsConnected = useRef<boolean | null>(false);
-  const { isConnected, loading: retryLoading, retryCheck } = useNetworkStatus();
+  const {
+    isConnected,
+    serverConnected,
+    loading: retryLoading,
+    retryCheck,
+  } = useNetworkStatus();
   const invalidateCache = useInvalidatePlaybackProgressCache();
   useEffect(() => {
     // Only invalidate cache when transitioning from offline to online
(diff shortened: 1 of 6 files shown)


components/settings/HomeIndex.tsx
```diff
@@ -358,13 +363,28 @@ export const HomeIndex = () => {
 
   const sections = settings?.home?.sections ? customSections : defaultSections;
 
-  if (isConnected === false) {
+  if (!isConnected || serverConnected !== true) {
```
Simon-Eklundh, line 366: Assuming isConnected is to the internet/similar:

If we aren't connected or we can't reach server: show offline message.

Is there ever a possibility that isConnected is false while serverconnected is true?

(like on lan with no internet)

providers/JellyfinProvider.tsx
```diff
@@ -374,14 +374,14 @@ function useProtectedRoute(user: UserDto | null, loaded = false) {
...
+    const inAuthGroup = segments.length > 1 && segments[0] === "(auth)";
 
     if (!user?.Id && inAuthGroup) {
       console.log("Redirected to login");
       router.replace("/login");
     } else if (user?.Id && !inAuthGroup) {
-      console.log("Redirected to home");
-      router.replace("/(auth)/(tabs)/(home)/");
+      console.log("Redirected to home", segments[0], !inAuthGroup);
+      // router.replace("/(auth)/(tabs)/(home)/");
```
Simon-Eklundh: why is this commented out?

this being commented out breaks the login flow.

Simon-Eklundh (commented):
really nice changes, but one question and one bug I've seen that needs looking over before I'm happy :)

components/settings/HomeIndex.tsx
lancechant: So is connected is now linked to just having a connection to the device, the old logic used connected to a network and internet reachable, so with this logic it should be the follows:

isConnected -> true, but no internet, you can then still reach the server and check for the server connection
Is connected -> false, you are not on any connection so can't connect to anything

Simon-Eklundh: Wouldn't serverconnected be the same as isConnected then?

lancechant: Uhm no, serverConnected checks that the server is up and reachable. IsConnected is just if you have network

Simon-Eklundh: Ahhh

I thought it was isConnected to each other :D

So to make sure my dumbass understands:

ServerConnected: client can reach server

isConnected: client has a network connection?

lancechant: Correcto, you can check the `useNetworkStatus` file to see the logic if you haven't already 😁

Simon-Eklundh: Then assuming the router is on but can't connect to the internet, and the user is on local network as well as the server, we'd have fully working offline-playback without needing to have downloaded from how I'm reading it 

Assuming ofc the phone is connected to the server on lan

lancechant: Yeah that is the idea/plan

Simon-Eklundh: Alright, I'll figure out how to resolve this on GitHub, and approve :)

Simon-Eklundh (approved):
been running this for a long time, it works. 

it does blip in with the offline screen on occasion when opening the app but I think that's fine

lostb1t (approved):

Issue #1165, opened 2025-11-07
[Bug]:  Fix the "hero" element, with device orientation and different sized posters


lancechant:
### This issue respects the following points:

- [x] This is a **bug**, not a question or configuration issue; please consult our community channels before filing a report. [Discord](https://discord.streamyfin.app).
- [x] This issue is **not** already reported on [GitHub](https://github.com/streamyfin/streamyfin/issues?q=is%3Aissue+is%3Aopen+label%3A"🐛%20bug") *(I've searched it)*.
- [x] I'm using an up-to-date version of Streamyfin. We generally do not support older versions. If possible, please update to the latest version before opening an issue.
- [x] I agree to follow Streamyfin's [Contribution rules](https://github.com/streamyfin/streamyfin/blob/develop/.github/CONTRIBUTING.md).
- [x] This report addresses only a single issue; If you encounter multiple issues, please create separate reports for each one.

### What happened?

Problem number 1. When you have different size posters, the size of the poster affects the size of the hero element. 

Problem 2: on ipadOS at least, changing orientation on the home screen doesn't reload the hero element properly, so it stays the same size it was in the original orientation. This leads to visual bugs. For example the hero showing both current and half of next item 

<img width="1329" height="873" alt="Image" src="https://github.com/user-attachments/assets/b43db70d-6ac2-4ab9-b214-f1484ffbdf65" />

### What did you expect to happen?

To keep a uniformed size, and keep same aspect on orientation.

### Reproduction steps

Have different sized posters?
Change device from landscape to portrait and back (possibly only on bigger devices)

### Which device and operating system are you using?

Any

### Streamyfin Version

TestFlight/Development build

### Relevant logs (if available)

```shell

```

PR #1166, opened 2025-11-07
fix: improve carousel behavior and settings option

fredrikburmester:
- Change large carousel to use same paralax as in item details. 
- Add option to disable in settings.


lancechant (dismissed):

lancechant:
related to #1165

lancechant (dismissed):

lancechant (approved):

PR #1174, opened 2025-11-09
feat: Expo 54 (new arch) support + new in-house download module 

fredrikburmester:


package.json
```diff
@@ -22,117 +22,110 @@
...
     "lodash": "^4.17.21",
     "nativewind": "^2.0.11",
-    "react": "19.0.0",
-    "react-dom": "19.0.0",
+    "patch-package": "^8.0.0",
+    "react": "19.1.0",
+    "react-dom": "19.1.0",
     "react-i18next": "^15.4.0",
-    "react-native": "npm:react-native-tvos@0.79.5-0",
+    "react-native": "0.81.5",
```
lancechant: removing TV?

fredrikburmester: There will be better TV support built into Expo 54 (I believe). So no we won't be dropping TV. But I also haven't tested TV with this build yet.

lancechant: But we can still use that repo? it shouldn't be much of an impact

fredrikburmester: We can't use react native `0.79.5` with Expo 54. We need `0.81.5`.

lancechant: It does have newer versions, [82.1.0](https://github.com/react-native-tvos/react-native-tvos/releases/tag/v0.82.1-0) their releases are tied to the react-native ones mostly

fredrikburmester: Ok good, then we can use https://github.com/react-native-tvos/react-native-tvos/releases/tag/v0.81.5-1

app/_layout.tsx
```diff
@@ -255,7 +255,7 @@ function Layout() {
 
     // only create push token for real devices (pointless for emulators)
     if (Device.isDevice) {
-      Notifications?.getExpoPushTokenAsync({
+      Notifications?.getDevicePushTokenAsync({
```
lancechant: Hey @fredrikburmester this doesn't take any arguments. https://docs.expo.dev/versions/latest/sdk/notifications/#getdevicepushtokenasync

Might need to add types to help with this, as the .then needs to change as well

fredrikburmester: Ah thanks for this.

package.json
fredrikburmester: @lancechant added back react native tv and it's building fine 👍🏻

PR #1199, opened 2025-11-15
fix: remove renovate, created too many prs, bad for expo projects

fredrikburmester:
Expo usually uses not the latest packages, that means renovate will just bloat our PR/Issue page.


fredrikburmester:
decided i'll close prs not wanted instead

Issue #883, opened 2025-07-27
[Bug]: Skip Credits Button not Working


RayJW:
### What happened?

As the title says, the skip credit button, if shown, does not seem to be doing anything for some media. It seems to work for reliably for some shows, and reliably not for others. Sometimes it also breaks the app until force closed by leaving the app frozen on the last second of the video. I'm not sure if it might be linked to skipping to the very end of the video not being handled correclty.

### Reproduction steps

1. Watch any media with the credits marked.
2. At the end of the video, the skip credits button will correctly show.
3. Press the button and observe, as nothing happens.

### Which device and operating system are you using?

iPhone 13 Pro Max, iOS 18.5

### Version

0.28.0

RayJW:
Still relevant.

irfanhakim-as:
In my case, the 'Skip Credits' button has never worked/done anything for me in any shows, on the Streamyfin client (Android and iOS). The 'Skip Intro' button works as intended.

PR #1277, opened 2025-12-18
fix(player): Fix skip credits seeking past video end causing pause

### LINKED ISSUE #172: Next episode does not play next episode
**Describe the bug**
In 0.17.0, when an episode ends, the player attempts to play the next episode. But it's not actually the next episode that's playing but the same that finished. Only the title text is updated to the next episode. 

**To Reproduce**
1. Play episode to the end
2. Pay attention which episode starts playing then

**Smartphone (please complete the following information):**
 - Device: Samsung Galaxy Tab A8 
 - OS: Android 14
 - Version 0.17.0 (GitHub release APK)

retrozenith:
## Problem

When pressing 'Skip Credits', the video would pause instead of seeking to the end of the credits. This made the skip credits feature unusable.

## Root Cause

The credit segment timestamps from Jellyfin can report an end time that slightly exceeds the actual video duration. When VLC tries to seek past the video end, it just pauses instead of playing.

## Solution

- Added `totalDuration` parameter to `useCreditSkipper` hook
- Check if seek target would go past the end and cap it to 2 seconds before video ends
- This skips the credits properly and triggers the normal next-episode flow

## Testing

Tested on an iOS emulator running on a macOS VM. Would appreciate verification on physical hardware.

## Changes

- `hooks/useCreditSkipper.ts` - Added duration check and seek position capping
- `components/video-player/controls/Controls.tsx` - Pass total duration to the hook

Fixes #172

### DIFF
--- a/components/video-player/controls/BottomControls.tsx
+++ b/components/video-player/controls/BottomControls.tsx
@@ -21,6 +21,7 @@ interface BottomControlsProps {
   isVlc: boolean;
   showSkipButton: boolean;
   showSkipCreditButton: boolean;
+  hasContentAfterCredits: boolean;
   skipIntro: () => void;
   skipCredit: () => void;
   nextItem?: BaseItemDto | null;
@@ -69,6 +70,7 @@ export const BottomControls: FC<BottomControlsProps> = ({
   isVlc,
   showSkipButton,
   showSkipCreditButton,
+  hasContentAfterCredits,
   skipIntro,
   skipCredit,
   nextItem,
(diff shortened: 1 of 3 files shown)


Simon-Eklundh (approved):
lgtm, tested on physical android device

RayJW:
I think this should also fix #883 as it sounds like what the behaviour seems to be for me.

retrozenith:
If description of your issue is accurately it could really be the fix.

fredrikburmester (approved):

PR #1299, opened 2025-12-30
feat: add chapter support

CeruleanRed:
# 📦 Pull Request

## 🔖 Summary
Added basic play bar markers for chapters and two new buttons to help navigate between them

## 🛠️ What’s Changed
- Type: feat
- Scope: player
- Add chapter support, with play bar markings and new player buttons

### ⚠️ Breaking Changes
None

None

### ⚡ Performance Impact
None

Screenshot_1767092558

Screenshot_1767092566

Screenshot_1767092572

## ✅ Checklist

## 🔍 Testing Instructions
1. Checkout this PR
2. Start any file that has chapters
3. Play around with chapter controls

## ⚙️ Deployment Notes
None

## 📝 Additional Notes
None

### DIFF
--- a/components/video-player/controls/BottomControls.tsx
+++ b/components/video-player/controls/BottomControls.tsx
@@ -195,7 +195,7 @@ export const BottomControls: FC<BottomControlsProps> = ({
                   key={i}
                   style={{
                     position: "absolute",
-                    backgroundColor: `#${i}FFF${i}${i}`,
+                    backgroundColor: "white",
                     width: 2,
                     height: "50%",
                     top: "25%",
--- a/components/video-player/controls/BottomControls.tsx
+++ b/components/video-player/controls/BottomControls.tsx
@@ -21,6 +21,7 @@ interface BottomControlsProps {
   showSkipButton: boolean;
   showSkipCreditButton: boolean;
   hasContentAfterCredits: boolean;
+  chapterStartMs?: number[];
   skipIntro: () => void;
   skipCredit: () => void;
   nextItem?: BaseItemDto | null;
@@ -69,6 +70,7 @@ export const BottomControls: FC<BottomControlsProps> = ({
   showSkipButton,
   showSkipCreditButton,
   hasContentAfterCredits,
+  chapterStartMs = [],
   skipIntro,
   skipCredit,
   nextItem,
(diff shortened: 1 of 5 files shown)


CeruleanRed:
UI and chapter markers on the play bar just serve as POCs atm, since I wasn't sure what the design should be

fredrikburmester:
Looks nice; maybe keep the tick marks white instead of green. How did you place them, absolutely, or does the slider support those ticks/marks/bars?

CeruleanRed:
The play bar itself supports ticks, however only at fixed intervals. I made the ticks basically by wrapping the play bar in an invisible view and positioning the ticks absolutely within it.

I think making the ticks white would make them invisible once the video has passed them, every implementation I've seen picked colors in such a way as to always keep them visible. The play bar library uses squares tilted 45°, maybe that would look cooler than lines.

CeruleanRed:
Alternatively to the "skip chapter" buttons, findroid uses long presses on either side to do chapter navigation. I find that's potentially something to consider since the buttons feel a bit crowded in the vertical layout.

fredrikburmester:
I agree with the "next chapter" buttons it becomes a bit crowded. Maybe we can make them smaller, and without the border? Also green is maybe not my favorite color, but hey it looks like it's working. What do you think? I think this is a nice addition, it just requires some tweaking.

CeruleanRed:
I can try making the icons smaller and the tick marks white. Probably on the weekend and then we take it from there

CeruleanRed:
This is what it would look like currently with the tick marks being white, and the next chapter buttons a bit smaller. I can't say I'm a big fan. I personally feel that the tick marks should be visible regardless of viewed status, so that one can see them and jump back if so desired. Having some of the buttons smaller is also not sparking joy as it feels like the design is inconsistent. Maybe there's a way to make all the buttons smaller in the vertical view and have them larger in the horizontal one?

Screenshot_1768735082

fredrikburmester:
Maybe you could make the tick marks taller, extending outside of the seek bar (both above and below), i think some players do this.

CeruleanRed:
> Maybe you could make the tick marks taller, extending outside of the seek bar (both above and below), i think some players do this.

I absolutely can, but if a chapter happens to be at the beginning or end of the file, that screws with the rounded corners of the player bar

fredrikburmester:
Can you investigate how other players do this, take some screenshots and compare?

CeruleanRed:
So here's what I found

VLC
image

Jellyfin App Web Player
image

Findroid
image

The general solution seems to be to not have rounded edges for the player bar

CeruleanRed:
How should I proceed here @fredrikburmester?

Gauvino:
Thanks for your contribution a refactor of your PR was merge !

PR #1320, opened 2026-01-07
fix(ci): use default GITHUB_TOKEN for Crowdin workflow

Gauvino:
Replaces custom CROWDIN_GITHUB_TOKEN secret with the default GITHUB_TOKEN provided by GitHub Actions.

This simplifies the authentication setup by leveraging the automatically available token instead of requiring a separate secret to be configured.


Simon-Eklundh (approved):
lfg!

PR #1332, opened 2026-01-10
feat: MPV player for both Android and iOS with added HW decoding PiP (with subtitles)

fredrikburmester:
Cred to @Alexk2309 making all the hard work on the native modules.


Alexk2309 (dismissed):
Tested this personally looks good

Issue #570, opened 2025-02-27
[Bug]: iOS app: Not being able to get past splash screen when disconnected from the server / issue opening downloads


carbon0range:
### What happened?

When connected to the server the app works great but when disconnected from the server I can't get past the splash screen (screen with streamyfin logo) to get to my downloads.  
Edit: now it will open after the app sitting open for over 2 minutes.

I tried testing the download feature when connected to the server and the media is compatible with the device but it won't play when I first click on it.  I have to click out and then click the download again to be able to view it when connected to the server.

### Reproduction steps

Repeat Stuck on Splash Screen Bug
1. Open/connect to server through streamyfin.
2. Download whatevery content is desired.
3. Once finished close streamyfin and turn off wifi/cellular.  
4. Open streamyfin and it will sit on the splash screen and won't move.

Repeat Opening Download when connected to server:
1. Download content.
2. When connected to server open downloads and play desired video.
3. Video will have loading spinner and will spin and spin endlessly.
4. Click out of video and then click it again and it plays instantly.

### Which device and operating system are you using?

iPhone 13, latest version of stable iOS.

### Version

0.27.0

KingKevin23:
I am having the same issue with the splash screen on iPadOS. Kinda poetic that I can not use the app/watch the downloaded media once I am offline

5quirrel:
Same issue

iOS 18.5

seanzi86:
I am having the same issue, stuck at the splash screen when I'm not connected to my home network. Is there any known work around?

Simon-Eklundh:
I just did some testing, I'd assume if you turn off your networking on the phone (mobile data) you should be able to access the off-line portions


(And then you should be able to turn on networking again, should just be at booting the app)

seanzi86:
I am on a WiFi only iPad running iOS 18.5, and when I disable WiFi, I get stuck at the splash screen. For now I have resorted to manually moving files to play.

cthulu:
Is this timeout something that can be made customizable, or at least shorter? I wasn't able to get to the downloads even after waiting long enough (but I didn't know up turn off my data completely)

Simon-Eklundh:
Idon't know for sure... Since for me it was pretty quick (other than when my server is off and I'm connecting through a reverse proxy 

But I'd assume this is a race condition somewhere....

yellowsnow3:
[...]

PR #1406, opened 2026-01-25
fix: Opening app in Offline mode (WiFI off or Server unrachable)

### LINKED ISSUE #1313: [Bug]: App freezes on splash screen when not connected to LAN network

### What happened?

I have my streamy fin app configured to point to a local host address (nas.local)
I am disconnected from my home network (mobile internet) and it just lingers on the splash screen with the streamyfin logo

I would expect there to be a screen where I can change the server address

### Reproduction steps

1. app is setup and working on home LAN with a local address
2. disconnect from wifi on my phone
3. open the app

iPhone 15 pro - latest OS

### Streamyfin Version

0.47.1

_No response_

### Screenshots or Videos

_No response_

```shell

```

### Additional information

_No response_

AleksanderEvensen:
# 📦 Pull Request

Motivation: The download feature from the Streamyfin app was the deciding feature for going with it in the first place, and oh boy was I in for a surprise when i tried it the first time. The app was endlessly loading on the splashscreen and wouldn't under any circumstances give me the option to go to my downloaded files (kinda ironic for a offline feature to not work while offline)

I did som digging and found the culprit to be in the getCurrentUser API fetch in the initializeJellyfin method.
https://github.com/streamyfin/streamyfin/blob/358e00d8b71f321bceb5575d5026979d52e8a449/providers/JellyfinProvider.tsx#L512

For some odd reason i dont know why, but it never fails / completes and ends in an infinite loop when WiFi is turned off on my apple devices (works well on my android device though).

So i tried modifying the SplashScreen hide logic to just hide the splashscreen imediately, and quickly realised you guys already handle status indication when fetching initial information from the jellyfin server. I therefore just kept the logic and it worked flawlessly, with and without a internet connection.

## 🔖 Summary

This just makes the Splashscreen instantly hide once the JellyfinProvider component is mounted.

## 🏷️ Ticket / Issue

Closes: #1313 #570 

## 🛠️ What’s Changed
Just one simple commit with the useEffect changes

### ⚠️ Breaking Changes
None

None

### ⚡ Performance Impact
None

**Before:**
https://github.com/user-attachments/assets/027c4dc6-a5f1-40d2-b50f-5e0839e7dedf

**After:**
https://github.com/user-attachments/assets/097704f6-6f0c-4251-8f6f-4ff0fde63069

## ✅ Checklist

- [ ] Docs updated (README/ADR/usage/API) <- No updating needed

## 🔍 Testing Instructions

1. Do the normal dev setup guides provided in the readme.
2. Start app with `bun run ios`
3. Connect to jellyfin server and download any media
4. Close app and turn off wifi on Macbook
6. Open app and observe the screen showing the "Go to downloads" button
^^^^^^^ This step didn't work before.

### DIFF
--- a/providers/JellyfinProvider.tsx
+++ b/providers/JellyfinProvider.tsx
@@ -559,10 +559,8 @@ export const JellyfinProvider: React.FC<{ children: ReactNode }> = ({
   };
 
   useEffect(() => {
-    if (loaded) {
-      SplashScreen.hideAsync();
-    }
-  }, [loaded]);
+    SplashScreen.hideAsync();
+  }, []);
 
   useProtectedRoute(user, loaded);
 


Simon-Eklundh (approved):
seems to work fine on android from my testing, but I'll defer to cagemaster just in case

lancechant:
So the main issue I have with it, is the authenticated logic and the `protectedRoute` logic at 
https://github.com/streamyfin/streamyfin/blob/358e00d8b71f321bceb5575d5026979d52e8a449/providers/JellyfinProvider.tsx#L583-L599

It currently doesn't do any checks for the download path at the moment, so this would work.. but maybe another option is to check if you offline, and check if the user is in the memory, since it's stored there, and then use that look here,
https://github.com/streamyfin/streamyfin/blob/358e00d8b71f321bceb5575d5026979d52e8a449/providers/JellyfinProvider.tsx#L604-L614 then it removes the call to that endpoint and should give the same results, but still with auth technically.. It doesn't seem to store expiry timestamps in that user, so could technically work for a while, while being offline.

This is just my opinion, but something I'd prefer. @AleksanderEvensen

AleksanderEvensen:
The current authentication logic is flawed imo. There should be a `_layout.tsx` file that wraps all routes under the `(auth)/` route. This layout should
1. Ensure the users wifi is enabled with `NetInfo`.
2. Ensure the server is actually reachable.
3. Ensure the "stored user"/current session actually has a connection to said jellyfin instance
4. Then and only then show the pages under `(auth)/`
   - If not then it should either redirect to the `/login` page or some other page that indicates to the user what is wrong

This should never be in a useEffect that is handled in a global context provider


For offline viewing there should be a route that is not nested under `(auth)/`. Essentially making the path structure somehting like:

```
/app 
 | - (auth)/
 |    | - home.tsx
 |    | - library.tsx
 |    | - ... etc ... (downloads.tsx should not be here)
 | - login.tsx
 | - downloads.tsx
``` 

Downloads page should be accessible regardless of authentication state and internet state.

Thoughts @lancechant?


Have also taken a deeper look into the `NetworkStatusProvider.tsx` and `JellyfinProvider.tsx`. There are lots of scenarios that can be refactored out into tanstack queries which will make the code base a lot more readable and give hooks that are easy to reuse. It can also essentially both Context Providers.

If you ask me all async state should be handled with a tanstack qurry because there are soooo many edge case scenarios that are missed when just doing a useEffect like this:

```ts
useEffect(async () => {
  setIsLoading(true)
  try {
    await fetchSomething(...dynamicParams)
  } catch(e) {
    setError(e)
  } finally {
  	setIsLoading(false)
  }
}, [...dynamicParams])
```

I will gladly take on the task to do this refactoring if you're okay with it?

lancechant:
Hey @AleksanderEvensen if you are willing to tackle it you are more than welcome to, the final decision is with @fredrikburmester if they have any reservations with the design. 

My only thought is the downloads do still need to be under Auth, because they are storing playback state, and need to be authenticated to do so and to sync back with the server when online so you at least need the user to store it against. 

I don't know much about the tan stacks but I'll look into them to learn more, but happy for you to tackle it and review the changes 😁👍

AleksanderEvensen:
> My only thought is the downloads do still need to be under Auth, because they are storing playback state, and need to be authenticated to do so and to sync back with the server when online so you at least need the user to store it against.

Ah i see makes sense.

> I don't know much about the tan stacks but I'll look into them to learn more, but happy for you to tackle it and review the changes 😁👍

If you want some excellent reading material on the topic i can recomend Tkdodo's blog series on react query especially [this one: Why you want react-query](https://tkdodo.eu/blog/why-you-want-react-query) which explains all the pitfalls that it handles for you.

> Hey @AleksanderEvensen if you are willing to tackle it you are more than welcome to, the final decision is with @fredrikburmester if they have any reservations with the design.

Ill wait for Fredrik's decision then before continuing. But it think i will still do some more work on this branch before then. Just some improvements to the already existing code.

lancechant:
Awesome I'll definitely give that a read 😃 and sounds good 😎

Gauvino:
This is completely fixed in https://github.com/streamyfin/streamyfin/pull/1723

PR #1484, opened 2026-03-10
fix(jellyseerr): auto re-login when Seerr session expires

### LINKED ISSUE #711: [Bug]: Jellyseerr login expirations are not properly handled
### What happened?

Jellyseerr cookies can expire for any number of reasons, and when that happens, there is no indication to the user what happened. The discover tab simply refuses to load, and there's not a clear way to re log into Jellyseerr. The only way to do this is to reset the Jellyseerr config and re type the server url AND password.

I propose that there should be pop up asking you to re enter the password when the cookie is expired 

Image
Image

### Reproduction steps

Attempt to use the discover tab with an invalid cookie.

Cookies will automatically invalidate after a while (1+ month), but as to specifically invalidate them, I'm not aware of a method in Jellyseerr. Could most likely delete the cookies from the database itself for debug, or just attempt to log in with an invalid cookie.

Android, S23 Ultra

### Version

0.28.0

_No response_

matschi95:
# 📦 Pull Request

## 🔖 Summary

Seerr sessions expire after 30 days. Previously this caused the Discover tab to silently show nothing, with no error message or feedback for the user. This PR implements automatic silent re-login when a 401/403 is detected, so the session is renewed transparently in the background without any user interaction. 

## 🏷️ Ticket / Issue

- Closes #711

## 🛠️ What's Changed

- Type: fix
- Scope: jellyseerr
- Short summary: Store Seerr credentials securely and auto re-login when the session expires, retrying all failed requests transparently.

### Credential storage

On every successful Seerr login — both manual (via Settings → Plugins → Jellyseerr) and automatic (triggered by Jellyfin login when a Seerr server URL is configured via the Streamyfin plugin) — the Jellyfin username and password are now persisted in `expo-secure-store` under a dedicated key (`jellyseerr_credentials`). Credentials are removed when the user disconnects from Seerr.

### Automatic re-login on 401 / 403

The Axios response interceptor in `JellyseerrApi` now catches 401 and 403 responses and attempts a silent re-login using the stored credentials before retrying the original request:

- A `static isRefreshing` flag and a request queue (`static queue`) on `JellyseerrApi` prevent multiple concurrent re-login attempts across all hook instances. The first 401 triggers the re-login; all other simultaneous requests are queued and retried automatically once the new session is established.
- If re-login succeeds the user notices nothing — content loads as normal.
- If re-login fails (no stored credentials, wrong password, server unreachable) all queued requests are rejected, the stale session data is cleared, and a single "Session Expired" toast is shown.

### Race condition fix

The relogin callback is registered inside `useMemo` (previously `useEffect`) so it is available before React Query fires its first requests. Using `useEffect` caused a race condition where the callback was still `null` when the initial 401 arrived.

### DEV testing button (`__DEV__` only)

A purple **"DEV: Simulate expired session"** button is added to the Seerr settings page. It:
1. Clears the native HTTP cookie cache via `@react-native-cookies/cookies` (`CookieManager.clearAll()`) so the server no longer recognises the session.
2. Corrupts the MMKV cookie entry so it remains non-empty (keeping `jellyseerrApi` alive) but invalid.

The button is rendered **outside** the `DisabledSetting` wrapper so it stays interactive even when the server URL is locked by the Streamyfin plugin.

### ⚠️ Breaking Changes

None.

The Jellyfin password is stored in `expo-secure-store`, which is encrypted on-device (iOS Keychain / Android Keystore). It is used exclusively for re-authenticating with Seerr and is never logged or transmitted elsewhere. This is the same mechanism already used for Jellyfin account tokens in this codebase.

### ⚡ Performance Impact

None for the happy path. On session expiry there is one additional login request before the original requests are retried.

No visible UI change in production. The DEV button is only visible in development builds.

## ✅ Checklist

- [x] Code follows project style and passes lint/format (`bun run lint`)

## 🔍 Testing Instructions

1. `git fetch origin pull/<PR_ID>/head:fix_seerr_cookie && git checkout fix_seerr_cookie`
2. Install deps: `bun install`
3. Build: `bun run prebuild && npm run android` (or `bun run ios`)
4. Log in to Seerr via **Settings → Plugins → Jellyseerr** (re-login required to store credentials).
5. Open the Jellyseerr settings page and tap **DEV: Simulate expired session**.
6. Navigate to **Search → Discover tab**.
7. Verification:
   - [ ] Console shows `[Seerr] Session expired, attempting re-login…`
   - [ ] Console shows `[Seerr] Re-login successful, retrying queued requests`
   - [ ] Discover tab loads normally with no user interaction
   - [ ] Only one re-login attempt despite multiple parallel requests
8. To test the failure path: tap the DEV button, then disconnect from Seerr (to remove stored credentials), then open Discover.
   - [ ] Console shows `[Seerr] Re-login failed` and `[Seerr] Session expired and re-login not possible, clearing session data`
   - [ ] A single "Session Expired" toast appears
   - [ ] No repeated toasts from React Query retries

## ⚙️ Deployment Notes

- Requires a native rebuild (`bun run prebuild`) due to the new `@react-native-cookies/cookies` dependency.
- Existing users who were already logged into Seerr before this update will not have credentials stored yet. They will need to re-login once via Settings → Plugins → Jellyseerr after updating. Subsequent session expirations will then be handled automatically.

## 📝 Additional Notes

- This implementation was developed with the assistance of [Claude Code](https://claude.ai/code) (Anthropic). 
- The `@react-native-cookies/cookies` package is used exclusively in the `__DEV__` testing button. It is not part of the production re-login flow.
- The re-login mechanism works for both manual Seerr logins and the auto-login triggered by the Streamyfin Jellyfin plugin.
- @herrrta is working on an alternative authentication approach that uses the Jellyfin Access Token instead of session cookies. However, this requires a change in the Seerr core (see [seerr-team/seerr#2244](https://github.com/seerr-team/seerr/pull/2244)) and it is unclear whether or when that change will be accepted. This PR provides a reliable interim fix using the existing cookie-based session flow.

### DIFF
--- a/app/(auth)/(tabs)/(home)/settings/plugins/jellyseerr/page.tsx
+++ b/app/(auth)/(tabs)/(home)/settings/plugins/jellyseerr/page.tsx
@@ -1,8 +1,12 @@
-import { ScrollView } from "react-native";
+import CookieManager from "@react-native-cookies/cookies";
+import { ScrollView, View } from "react-native";
 import { useSafeAreaInsets } from "react-native-safe-area-context";
+import { toast } from "sonner-native";
+import { Button } from "@/components/Button";
 import DisabledSetting from "@/components/settings/DisabledSetting";
 import { JellyseerrSettings } from "@/components/settings/Jellyseerr";
 import { useSettings } from "@/utils/atoms/settings";
+import { storage } from "@/utils/mmkv";
 
 export default function page() {
   const { pluginSettings } = useSettings();
@@ -22,6 +26,21 @@ export default function page() {
       >
         <JellyseerrSettings />
       </DisabledSetting>
+      {__DEV__ && (
+        <View className='px-4 pt-2'>
+          <Button
+            color='purple'
+            onPress={async () => {
+              await CookieManager.clearAll();
+              // Keep MMKV cookies non-empty so jellyseerrApi stays alive
+              storage.setAny("JELLYSEERR_COOKIES", ["XSRF-TOKEN=invalid"]);
+              toast.info("DEV: Session invalidated – open Discover");
+            }}
+          >
+            DEV: Simulate expired session
+          </Button>
+        </View>
+      )}
     </ScrollView>
   );
 }
--- a/app/(auth)/(tabs)/(home)/settings/plugins/jellyseerr/page.tsx
+++ b/app/(auth)/(tabs)/(home)/settings/plugins/jellyseerr/page.tsx
@@ -1,12 +1,8 @@
-import CookieManager from "@react-native-cookies/cookies";
-import { ScrollView, View } from "react-native";
+import { ScrollView } from "react-native";
 import { useSafeAreaInsets } from "react-native-safe-area-context";
-import { toast } from "sonner-native";
-import { Button } from "@/components/Button";
 import DisabledSetting from "@/components/settings/DisabledSetting";
 import { JellyseerrSettings } from "@/components/settings/Jellyseerr";
 import { useSettings } from "@/utils/atoms/settings";
-import { storage } from "@/utils/mmkv";
 
 export default function page() {
   const { pluginSettings } = useSettings();
@@ -26,21 +22,6 @@ export default function page() {
       >
         <JellyseerrSettings />
       </DisabledSetting>
-      {__DEV__ && (
-        <View className='px-4 pt-2'>
-          <Button
-            color='purple'
-            onPress={async () => {
-              await CookieManager.clearAll();
-              // Keep MMKV cookies non-empty so jellyseerrApi stays alive
-              storage.setAny("JELLYSEERR_COOKIES", ["XSRF-TOKEN=invalid"]);
-              toast.info("DEV: Session invalidated – open Discover");
-            }}
-          >
-            DEV: Simulate expired session
-          </Button>
-        </View>
-      )}
     </ScrollView>
   );
 }
--- a/hooks/useJellyseerr.ts
+++ b/hooks/useJellyseerr.ts
@@ -1,5 +1,4 @@
 import axios, { type AxiosError, type AxiosInstance } from "axios";
-import * as SecureStore from "expo-secure-store";
 import { atom } from "jotai";
 import { useAtom } from "jotai/index";
 import { inRange } from "lodash";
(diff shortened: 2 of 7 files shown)


app/(auth)/(tabs)/(home)/settings/plugins/jellyseerr/page.tsx
```diff
@@ -22,6 +26,21 @@ export default function page() {
       >
         <JellyseerrSettings />
       </DisabledSetting>
+      {__DEV__ && (
+        <View className='px-4 pt-2'>
+          <Button
+            color='purple'
+            onPress={async () => {
+              await CookieManager.clearAll();
```
lancechant: The only usage is here, so I would say remove the CookieManager + package

hooks/useJellyseerr.ts
```diff
@@ -64,6 +65,26 @@ interface SearchResults {
...
 const JELLYSEERR_COOKIES = "JELLYSEERR_COOKIES";
+const JELLYSEERR_CREDENTIALS_KEY = "jellyseerr_credentials";
+
+interface JellyseerrCredentials {
+  username: string;
+  password: string;
+}
+
+export const saveJellyseerrCredentials = async (
+  username: string,
```
lancechant: So what are the chances the jellyseer and jellyfin user are different? why not just get the user auth and re-auth from that? As the app already stores it no?

lancechant (commented):
Thanks for getting around to it, there was talk with seerr that would make this redundant hence why it wasn't put in earlier, not sure on the status of that though, Have given some comments

hooks/useJellyseerr.ts
matschi95: Good point – the app already forces the same Jellyfin user for Seerr login (no username field, just a password), so storing separate credentials was redundant. The existing ServerCredential only stores the tokens and session cookie, but not the password – which is needed for Seerr re-authentication. So I added an optional seerrPassword field to ServerCredential instead of maintaining a separate store. It gets saved both during Jellyfin login and when manually logging into Seerr via settings (for QuickConnect users who don't have a password stored). The Seerr re-login callback now reads from there instead of its own SecureStore entry.

app/(auth)/(tabs)/(home)/settings/plugins/jellyseerr/page.tsx
matschi95: Done – removed the DEV button and the @react-native-cookies/cookies dependency entirely. The button was only there for testing the re-login flow, since I couldn't find a quick way to manually invalidate the Seerr session cookie otherwise.

hooks/useJellyseerr.ts
lancechant: Why does it read like an Ai replied here?🤣🤦‍♂️

lancechant (approved):

lancechant:
Do note, this might not get merged.. since as mentioned there is a seerr logic request that should not need this

PR #1492, opened 2026-03-14
fix(navigation): prevent duplicate screen pushes on rapid taps

a-collado:
# 📦 Pull Request

## 🔖 Summary

Prevent duplicate screens from being pushed when the user taps a navigation 
element multiple times rapidly, particularly on slow devices.

## 🛠️ What’s Changed
- Type: fix |
- Scope: navigation
- Short summary: `useAppRouter`: add a 300ms debounce guard on `push` to silently discard 
  duplicate calls within the same navigation action. `replace` is intentionally 
  excluded as replacing the current screen is always a deliberate action.

## 📋 Details

On slow Android devices, tapping a navigation element quickly could trigger 
multiple `push` calls before the first screen had finished rendering, resulting 
in duplicate screens on the navigation stack.

The fix adds a `isNavigatingRef` flag that blocks any additional `push` calls 
for 500ms after the first one. This window is long enough to cover rendering 
time on low-end hardware, but short enough to be imperceptible as a limitation 
to the user.

`replace` does not need the guard — replacing the current screen is always 
intentional and semantically different from pushing a new one.

Before:

https://github.com/user-attachments/assets/bb0b27bf-b974-4679-920e-522290cf5fa0

After:

https://github.com/user-attachments/assets/48956f08-d9d1-412b-8b58-bd91a17a9f70

## ✅ Checklist

## 🔍 Testing Instructions

1. Tap the same item multiple times rapidly
2. Verify only one screen is pushed, no duplicates on the stack
3. Verify that `replace`-based navigation (e.g. player screen changes) 
   still works normally

## 📝 Additional Notes

300ms was chosen as the debounce window, long enough to cover slow device 
rendering, short enough to be unnoticeable. Can be adjusted via the 
`NAVIGATION_DEBOUNCE_MS` constant in `hooks/useAppRouter.ts`.

### DIFF
--- a/hooks/useAppRouter.ts
+++ b/hooks/useAppRouter.ts
@@ -1,6 +1,8 @@
 import { useRouter } from "expo-router";
-import { useCallback, useMemo } from "react";
+import { useCallback, useMemo, useRef } from "react";
 import { useOfflineMode } from "@/providers/OfflineModeProvider";
+
+const NAVIGATION_DEBOUNCE_MS = 500;
 
 /**
  * Drop-in replacement for expo-router's useRouter that automatically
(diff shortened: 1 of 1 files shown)


lancechant (approved):

a-collado:
Hi, any news on this PR getting approved? Thanks!

Gauvino:
Hello, thank for the bug fix but i didn't like it this way, the fix works, but personally, I like the fixed timer only moderately.
A hard delay always ends up being too short on slow hardware or too long on fast one. Adjusting the screen focus prevents guessing a duration.

PR #1490, opened 2026-03-14
fix(player): restore playback position for downloaded files on Android

a-collado:
# 📦 Pull Request

## 🔖 Summary
Playback always starting from the beginning instead of 
resuming from the saved position. Affected Android physical devices only 
— not reproducible on emulator.

## 🛠️ What’s Changed

- Type: fix 
- Scope: player
- Short summary: Playback of downloaded files always started from the beginning instead of resuming from the saved position.

## 📋 Details

- fix(player): use `setPropertyDouble("time-pos")` instead of seek 
  command in Android MPV module — seek commands are silently ignored 
  by MPV during player initialization

- fix(player): prevent `videoSource` from recalculating `startPosition` 
  from `playbackPositionFromUrl` changes during playback

### 🖼️ Screenshots / GIFs (if UI)

Before:

https://github.com/user-attachments/assets/6d74abe1-bd49-4f62-aff1-4082b3b0977d

After:

https://github.com/user-attachments/assets/5ea836a0-be9d-4884-b477-3b19b4248658

## ✅ Checklist

- [x] I’ve read the [contribution guidelines](CONTRIBUTING.md)
- [x] Code follows project style and passes lint/format (`npm|pnpm|yarn|bun` scripts)
- [x ] Type checks pass (tsc/biome/etc.)
- [ ] Docs updated (README/ADR/usage/API)
- [x] No secrets/credentials included; env vars documented
- [ ] Release notes/CHANGELOG entry added (if applicable)
- [x] Verified locally that changes behave as expected

## 🔍 Testing Instructions

Tested on: Nothing Phone 2a Android 16
Not reproducible on Android emulator — requires physical device.

1. Download any media item
2. Play it past the 5% mark, then exit
3. Re-open and play using "Play downloaded file"
4. Verify playback resumes from the correct position

PR #1523, opened 2026-04-01
fix(player): restore offline playback resume progress

jevonearth:
## Summary

This fixes offline playback resume progress not being restored reliably after closing and reopening playback.

I hit this on a flight, while using the app the way it is meant to be used, with downloaded media in airplane mode. It was a real annoyance, and enough of a bummer that I dug in and fixed it.

## What changed

- use the downloaded item's local `PlaybackPositionTicks` when launching offline playback
- persist offline playback progress one final time when stopping playback
- invalidate local item, episode, adjacent, and series queries after local playback updates so resume state refreshes correctly

## Testing

- verified manually on a real Pixel 10 device
- tested with real downloaded media playback in airplane mode
- confirmed downloaded media now resumes from the saved position after pausing, closing the app, and reopening it
- ran:

```bash
./node_modules/.bin/biome check hooks/usePlaybackManager.ts components/PlayButton.tsx 'app/(auth)/player/direct-player.tsx'
```

## Notes

Full repo `typecheck` was not used as signoff because the repo currently has unrelated pre-existing Jellyseerr type errors outside this change set.

## AI Assistance Disclosure

This PR was prepared with AI assistance. I reviewed the code changes and tested the fix manually on-device.

### DIFF
--- a/app/(auth)/player/direct-player.tsx
+++ b/app/(auth)/player/direct-player.tsx
@@ -381,24 +381,6 @@ export default function page() {
     revalidateProgressCache,
   ]);
 
-  const stop = useCallback(() => {
-    // Update URL with final playback position before stopping
-    router.setParams({
-      playbackPosition: msToTicks(progress.get()).toString(),
-    });
-    reportPlaybackStopped();
-    setIsPlaybackStopped(true);
-    videoRef.current?.pause();
-    revalidateProgressCache();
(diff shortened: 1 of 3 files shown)


bbijsterbosch:
I believe this has already been fixed in https://github.com/streamyfin/streamyfin/pull/1490

lancechant (approved):

Gauvino:
Already fixed in develop

Issue #1298, opened 2025-12-29
[Bug]: Sometimes download would get stuck on Android


fishy:
### This issue respects the following points:

- [x] This is a **bug**, not a question or configuration issue; please consult our community channels before filing a report. [Discord](https://discord.streamyfin.app).
- [x] This issue is **not** already reported on [GitHub](https://github.com/streamyfin/streamyfin/issues?q=is%3Aissue+is%3Aopen+label%3A"🐛%20bug") *(I've searched it)*.
- [x] I'm using an up-to-date version of Streamyfin. We generally do not support older versions. If possible, please update to the latest version before opening an issue.
- [x] I agree to follow Streamyfin's [Contribution rules](https://github.com/streamyfin/streamyfin/blob/develop/.github/CONTRIBUTING.md).
- [x] This report addresses only a single issue; If you encounter multiple issues, please create separate reports for each one.

### What happened?

When downloading videos (the cloud icon in the UI) on Android, sometimes the download would proceed without any issue, but some videos would just get stuck at 0% forever and eventually fail/timeout. This happens in about 1 in 5 or maybe 1 in 10 downloads.

The same video (that got stuck) would download fine in Fladder Android version.

### What did you expect to happen?

Download should always proceed without issue, or give me a clear message on why it cannot be downloaded.

### Reproduction steps

1. Open Streamyfin Android
2. Go to a video (movie/episode/etc.), click the cloud icon to download it locally
3. Go back to home screen of Streamyfin
4. Click the downloads icon left to "Home"
5. The download stucks at 0% without any progress

### Which device and operating system are you using?

Pixel 10 Pro Fold with latest stable release of Android (I believe it's Android 16 QPR 2), Build number `BP4A.251205.006.E1`

### Streamyfin Version

0.47.1

### Jellyfin Server Information

- Jellyfin server: `10.11.5` via podman
- Local connection but with VPN on on the Android phone.

### Relevant logs (if available)

```shell

```

fishy:
Looking at server activity logs, the failed downloads don't even register on the server (there's no `UserDownloadingContent` event for the failed downloads). So my guess is that Streamyfin somehow failed to issue the download api to the server on those ones?

mazarian:
Just wanted to come here and say I'm having the exact same issue.

Issue #615, opened 2025-03-19
[Bug]: On-Device downloads abruptly stop upon screen time out


blobberbun:
### What happened?

When I queue a download to my device, I can see the download occurring in the app download section. However, if the screen goes to sleep (without closing or force-quitting the app), the video will show as fully downloaded.

However, upon playback from downloads the video will only play up to the percentage that the video was downloaded prior to the screen going to sleep.

### Reproduction steps

1. Queue up a video download  to device
2. allow phone screen to sleep
3. attempt to play video from downloads

### Which device and operating system are you using?

iPhone 14 Pro, iOS 18.3.2

### Version

0.27.0

adamzvolanek:
Seeing the same on my Pixel 7 (Android 15).

volrod64:
I think it's due to ram usage. I set my phone to never go to sleep by himself and let the app in first plan. After 10 minutes it crash

adamzvolanek:
Not stale, still an issue.

adamzvolanek:
Not stale, still an issue.

blobberbun:
Still an issue on iOS

PR #1546, opened 2026-04-24
fix(android): hold wake and wifi locks during downloads

thomasdf:
# 📦 Pull Request

## 🔖 Summary

Downloads die when you lock the screen. The foreground service holds no wake lock, so the CPU suspends during Doze, OkHttp's read timeout fires, and the download silently vanishes. This adds a `PARTIAL_WAKE_LOCK` so downloads survive screen-off.

## 🏷️ Ticket / Issue

- Related: #615
- Related: #1298

## 🛠️ What's Changed

- Type: **fix**
- Scope: **android/background-downloader**
- Short summary: Acquire `PARTIAL_WAKE_LOCK` in `DownloadService` lifecycle

## 📋 Details

`DownloadService` runs as a `dataSync` foreground service with a notification, but never acquires a `PowerManager.WakeLock`. When the screen locks, Android Doze suspends the CPU. The TCP connection stalls, OkHttp's 60-second `readTimeout` fires, the partial file gets deleted, and the download is just gone. Meanwhile Jellyfin keeps transcoding because it has no idea the client disconnected.

The fix acquires a `PARTIAL_WAKE_LOCK` in `onCreate()` and releases it in `onDestroy()`. The service already starts with the first download and stops when the queue empties, so the lock is held exactly when needed.

The original draft also held a `WifiLock` (`WIFI_MODE_FULL_HIGH_PERF`), but testing showed the wake lock alone is sufficient — a 15+ minute download through screen-off + USB disconnect completed cleanly with no WifiLock. The WifiLock has also been deprecated since API 34 (silently converted to `LOW_LATENCY`, which only activates with the screen on), so dropping it avoids the deprecation question entirely. See commit history for the iteration.

Two files, ~10 lines.

### ⚠️ Breaking Changes

None

None. `WAKE_LOCK` is a normal permission, no runtime prompt.

### ⚡ Performance Impact

The lock is held only while downloads are active with a visible notification. No battery impact when not downloading.

## ✅ Checklist

- [x] I've read the contribution guidelines
- [x] Code follows project style and passes lint/format
- [x] Type checks pass

## 🔍 Testing Instructions

1. Install on a physical Android device (emulator won't trigger Doze)
2. Start downloading a video that'll take a few minutes
3. Lock the screen, optionally unplug USB
4. Wait 5+ minutes
5. Unlock — download should still be going or finished

Logcat: `adb logcat -s DownloadService OkHttpDownloadManager`

Tested on a Tab S9+ — 15+ minute download survived screen lock + USB disconnect. Without the patch, the same download dies after about a minute.

## 📝 Additional Notes

I used Claude Code to help write this PR. Tested and verified on physical hardware (Tab S9+).

Longer term, migrating the download infrastructure to `WorkManager` would handle wake locks, retry, and constraints out of the box. Happy to do that as a follow-up if you'd prefer — this is just the minimal fix for the immediate problem.

## Summary by CodeRabbit

* **Improvements**
  * Background downloads now maintain device wake state during active transfers to ensure reliable completion without interruption from device sleep or power management.


modules/background-downloader/android/src/main/java/expo/modules/backgrounddownloader/DownloadService.kt
```diff
@@ -36,6 +40,17 @@ class DownloadService : Service() {
     super.onCreate()
     Log.d(TAG, "DownloadService created")
     createNotificationChannel()
+
+    val pm = getSystemService(PowerManager::class.java)
+    wakeLock = pm.newWakeLock(PowerManager.PARTIAL_WAKE_LOCK, "Streamyfin::DownloadWakeLock")
+    wakeLock?.acquire()
+
+    val wm = getSystemService(WifiManager::class.java)
+    @Suppress("DEPRECATION")
```
lancechant: When is this to be fully depreciated? What's the alternative?

thomasdf: Good catch — the deprecation comment prompted me to actually test whether the WifiLock was earning its keep. It turns out it wasn't, at least on the device I have to test with.

`WIFI_MODE_FULL_HIGH_PERF` was deprecated in API 34 (Android 14, Oct 2023) and on Android 14+ is silently converted to `WIFI_MODE_FULL_LOW_LATENCY` at the system level. `LOW_LATENCY` only activates with the screen on and the app foregrounded — exactly what's not true for our backgrounded downloads — so the WifiLock isn't really doing what the original code seemed to expect on modern Android.

I built a wake-lock-only variant and reran the same test on the Tab S9+ (Android 14): a 15+ minute download survived screen-off + USB disconnect with the WifiLock removed. Pushed that as a follow-up commit so the final state is just the `PARTIAL_WAKE_LOCK` — no deprecation suppression, simpler lifecycle. Let me know if you'd like me to squash before merge.

lancechant: Firstly @thomasdf This comment smells like an AI wrote it, why? secondly do you understand what's been done? If the AI is just writing things and you not validating it, and ensuring it is what's needed then I'm hesitant to even approve. I would rather you type your responses instead of using an LLM to respond. Unless it's a language thing, but then mention it, and give the original language you speak with, and a LLM translation

thomasdf: Smells like AI because it was AI. But ok, here I am - a real human being typing manually. The problem: When I use Streamyfin to download content, I have to keep the screen alive, and if not - within a short time after I lock the device, the download stops and just disappears from Streamyfin. At the same time, the server keeps downloading in the background. Jellyfin gets no info about the app giving up the download.

So, why does Streamyfin's download stop when I lock the device? From what I've found: okHttp gets killed when Android enters Doze and suspends the CPU, because the foreground service holds no wake lock to keep it running. i initially also thought that this was because Wifi went into low power mode, and not actually verifying which of these two things was the actual root cause is just sloppy.

So - why does aquiring the wakelock solve the problem? Because that makes Android keep the CPU running so OkHttp can keep reading bytes from the socket. How I've validated it? Built and installed the unmodified version on my Tab S9+, queued a download, locked the screen - waited about a minute, download dissapears before I've unlocked. Then, installed the modified version and re-tested. Now, the download keeps going.

Is this a good solution? Not really - but moving over to `WorkManager` is a major refactor.

lancechant: Appreciate that, and that makes a difference chatting to a bot response vs a human one, just saying 😅

Cool I understood the problem from the PR already, but thanks for clarifying it again here again, just wanted to ensure this was reviewed and understood by a human for maintainability. As might not be you looking at it again in the future.

If possible could you add a comment or something to say how moving over to the workManager would entail? So it is there for future work maybe? If you would like to do a PR for it then that's welcomed to. I'll approve this so long, but for when it will be merged that is unknown, as have to wait for other dev to be started up again and merged, so approving it will at least maybe put it on top for when that happens.

lancechant (approved):

Gauvino (approved):

PR #1558, opened 2026-05-17
fix: tv overlay focus navigation

stevebyatt10:
# 📦 Pull Request

## 🔖 Summary
Improve TV focus navigation for playback overlay actions, including skip intro/credits and next episode.

## 🏷️ Ticket / Issue
None.

## 🛠️ What’s Changed
- Type: fix
- Scope: tv
- Short summary: Adds TV focus guide destinations for right-aligned playback overlay cards so they can be reached from the control row.

## 📋 Details
_Details made by Codex_
The skip segment and next episode cards are positioned above the controls on the right side of the player. On tvOS, the geometric focus engine can fail to move focus from left-side controls, such as play/pause, to those overlay actions.

This adds explicit `TVFocusGuideView` navigation from the controls to the visible overlay card. When controls are visible, moving down from the overlay returns to play/pause. The code path is shared TV UI, so Android TV also receives the behavior, but this primarily addresses the tvOS focus issue.

### ⚠️ Breaking Changes
None.

None.

### ⚡ Performance Impact
None expected. This only adds focus refs and focus guide views around existing TV controls.

Not included. Change affects TV remote focus behavior.

## ✅ Checklist
- [x] Docs not required for this focus-only UI fix
- [x] Release notes/CHANGELOG entry not required

## 🔍 Testing Instructions
1. Run the TV app on Apple TV / tvOS.
2. Start playback for media with intro or credit segments.
3. Open the playback controls.
4. Focus the play/pause button or another control-row button.
5. Press up and verify focus moves to the visible skip intro/credits button.
6. Press down and verify focus returns to play/pause.
7. Near the end of an episode, verify the same up/down behavior with the next episode card.

Validation run:
- `npm run typecheck`
- `npx biome check components/tv/TVSkipSegmentCard.tsx components/tv/TVNextEpisodeCountdown.tsx components/video-player/controls/Controls.tv.tsx`

## ⚙️ Deployment Notes
No native rebuild required. This is a JS/TS focus-navigation change in the TV player controls.

## 📝 Additional Notes
The fix uses `TVFocusGuideView` because tvOS does not reliably support direct directional focus routing for this layout.

### DIFF
--- a/components/tv/TVNextEpisodeCountdown.tsx
+++ b/components/tv/TVNextEpisodeCountdown.tsx
@@ -40,7 +40,7 @@ export interface TVNextEpisodeCountdownProps {
   /** Whether this component should receive initial focus */
   hasTVPreferredFocus?: boolean;
   /** Destination used when moving down from this card */
-  downDestination?: RNView | null;
+  playButtonRef?: RNView | null;
 }
 
 // Position constants
@@ -57,7 +57,7 @@ export const TVNextEpisodeCountdown: FC<TVNextEpisodeCountdownProps> = ({
   controlsVisible = false,
   refSetter,
   hasTVPreferredFocus = true,
-  downDestination,
+  playButtonRef: downDestination,
 }) => {
   const typography = useScaledTVTypography();
   const { t } = useTranslation();
--- a/components/tv/TVNextEpisodeCountdown.tsx
+++ b/components/tv/TVNextEpisodeCountdown.tsx
@@ -7,7 +7,9 @@ import {
   Image,
   Pressable,
   Animated as RNAnimated,
+  type View as RNView,
   StyleSheet,
+  TVFocusGuideView,
   View,
 } from "react-native";
 import Animated, {
@@ -33,6 +35,12 @@ export interface TVNextEpisodeCountdownProps {
   onPlayNext?: () => void;
   /** Whether controls are visible - affects card position */
   controlsVisible?: boolean;
+  /** Callback ref setter for focus guide destination pattern */
+  refSetter?: (ref: RNView | null) => void;
+  /** Whether this component should receive initial focus */
+  hasTVPreferredFocus?: boolean;
+  /** Destination used when moving down from this card */
+  downDestination?: RNView | null;
 }
 
 // Position constants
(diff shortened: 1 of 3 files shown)


components/tv/TVNextEpisodeCountdown.tsx
```diff
@@ -172,6 +184,12 @@ export const TVNextEpisodeCountdown: FC<TVNextEpisodeCountdownProps> = ({
           </BlurView>
         </RNAnimated.View>
       </Pressable>
+      {downDestination && (
```
lancechant, line 187: @stevebyatt10 Can you explain how this solves the issue? as I see it's pretty much the playbutton ref or null passed in?

components/video-player/controls/Controls.tv.tsx
```diff
@@ -1040,6 +1044,9 @@ export const Controls: FC<Props> = ({
         onPress={skipIntro}
         type='intro'
         controlsVisible={showControls}
+        refSetter={setSkipSegmentRef}
+        hasTVPreferredFocus={!showControls}
+        downDestination={showControls ? playButtonRef : null}
```
lancechant: Depending on the answer above, I think this needs a rename

components/tv/TVNextEpisodeCountdown.tsx
stevebyatt10: Main fix is in `Controls.tv.tsx
`
```
  {/* Upward: control buttons → visible skip segment or next episode card */}
  {topOverlayFocusTarget && (
    <TVFocusGuideView
      destinations={[topOverlayFocusTarget]}
      style={styles.focusGuide}
    />
  )}

```
Which makes it so that the skip button or next episode button is focused from any button in the control bar, not having to be focused directly under it on the right side.

I was gonna store what button we came from so that when we went down it went back to that button, which is why `downDestination`. But felt it was a lot to store for not much benefit so just back down to play. I can update with that logic if we want.

lancechant: So If I read it correctly, up clicks on the remote would focus on the next/skip buttons right? but down would select play button? just ensuring I get the correct flow here.

No that's fine having the play selected, but then just I think have the variable renamed

stevebyatt10: Yeah correct, will rename it 🙏

PR #1559, opened 2026-05-17
fix: handle TV menu and back navigation

stevebyatt10:
# 📦 Pull Request

## 🔖 Summary
Fixes TV menu/back handling so player can use custom back button handling logic and root tab can exit to home screen.

## 🛠️ What’s Changed
- Type: fix
- Scope (optional): tvos, tv
- Short summary: simplify TV back/menu ownership so home root can exit natively and the player handles back/menu through a shared TV back hook.

## 📋 Details
_Generated by codex_

On the player screen pressing the back button was navigating back and not performing any of the custom logic like hiding controls or showing alert before exiting.

From some added logs:
```
  [player-back-handler] ...
  [tv-back-handler:event] {"eventType":"menu"}
  [tv-back-handler] {"segments":["(auth)","player","direct-player"]}
  [player-tv-event] {"eventType":"menu"}
  [player-back-handler] ...
```

Issue was overlapping ownership: the tab/home back handler was still mounted while the player route was active, and both layers could react to the same back event. That meant the tab handler could also navigate/back while the player was trying to run its custom logic.

- Renamed/simplified the tab-level handler to `useTVHomeBackHandler`.
   - `useTVHomeBackHandler` disables tvOS menu key interception on the home tab root so native tvOS app-exit behavior can run, and enables interception away from home root where JS-owned handling is needed.
   - Removed the custom route handling which is no longer needed and was preventing the player's custom back button handling
- Added `useTVBackPress`, a small wrapper around React Native `BackHandler` for TV platforms.
- Updated the player remote-control hook to use `useTVBackPress` for player exit behavior:
  - First back/menu press hides visible controls.
  - When controls are hidden, back/menu prompts before stopping playback.
- Kept `useTVEventHandler` focused on remote-control events like play/pause, D-pad, and long seek; it no longer owns menu/back behavior.

### ⚠️ Breaking Changes
None.

None. No data, permissions, credentials, or network behavior changed.

### ⚡ Performance Impact
None.

Not applicable. Behavior-only TV remote navigation fix.

## ✅ Checklist
- [x] Docs updated (README/ADR/usage/API)
- [x] Release notes/CHANGELOG entry added (if applicable)

## 🔍 Testing Instructions
1. Build/run the TV app.
2. On tvOS:
   - Open the home tab root and press the Apple TV Menu/Back button.
   - Verify the system/native app-exit behavior runs.
   - Start playback and press Menu/Back with controls visible.
   - Verify controls hide instead of navigating away.
   - Press Menu/Back again with controls hidden.
   - Verify the stop playback confirmation appears and stopping exits the player.
3. On Android TV:
   - Start playback and press hardware Back with controls visible.
   - Verify controls hide.
   - Press Back again with controls hidden.
   - Verify the stop playback confirmation appears and stopping exits the player.

### DIFF
--- a/components/video-player/controls/hooks/useRemoteControl.ts
+++ b/components/video-player/controls/hooks/useRemoteControl.ts
@@ -1,20 +1,7 @@
 import { useEffect, useRef, useState } from "react";
 import { Alert, BackHandler, Platform } from "react-native";
 import { type SharedValue, useSharedValue } from "react-native-reanimated";
-
-// TV event handler with fallback for non-TV platforms
-let useTVEventHandler: (callback: (evt: any) => void) => void;
-if (Platform.isTV) {
-  try {
-    useTVEventHandler = require("react-native").useTVEventHandler;
-  } catch {
-    // Fallback for non-TV platforms
-    useTVEventHandler = () => {};
-  }
-} else {
-  // No-op hook for non-TV platforms
-  useTVEventHandler = () => {};
-}
+import { useTVEventHandler } from "@/hooks/useTVEventHandler";
 
 interface UseRemoteControlProps {
   showControls: boolean;
--- a/components/video-player/controls/hooks/useRemoteControl.ts
+++ b/components/video-player/controls/hooks/useRemoteControl.ts
@@ -1,6 +1,7 @@
 import { useEffect, useRef, useState } from "react";
-import { Alert, BackHandler, Platform } from "react-native";
+import { Alert } from "react-native";
 import { type SharedValue, useSharedValue } from "react-native-reanimated";
+import { useTVBackPress } from "@/hooks/useTVBackPress";
 import { useTVEventHandler } from "@/hooks/useTVEventHandler";
 
 interface UseRemoteControlProps {
@@ -57,7 +58,6 @@ interface UseRemoteControlProps {
  */
 export function useRemoteControl({
   showControls,
-  toggleControls,
   togglePlay,
   onBack,
   onHideControls,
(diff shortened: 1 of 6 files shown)


patches/react-native-screens+4.18.0.patch
```diff
@@ -0,0 +1,86 @@
+diff --git a/node_modules/react-native-screens/ios/RNSScreenStack.mm b/node_modules/react-native-screens/ios/RNSScreenStack.mm
+--- a/node_modules/react-native-screens/ios/RNSScreenStack.mm
++++ b/node_modules/react-native-screens/ios/RNSScreenStack.mm
+@@ -34,6 +34,11 @@
```
lancechant: @stevebyatt10 how'd you get to this patch? is it going to be temp? because of future maintainability. Is there a PR to the repo to add this? What does it fix or try to achieve? I see it's adding the back and menu events, but why?

stevebyatt10: We weren’t getting tvOS back/menu events coming through to JS at all. I’m not sure how long that has been broken for, but without this patch our `BackHandler` / TV navigation logic never sees the menu press in some screens. So basically on tv player it just always navigates back, no controls hiding logic or confirm dialog.

Was debugging the JS side for quite a bit and couldn't get the event emitting. It looks like `react-native-screens`’ native stack was just not forwarding the event, so this patch forwards it through instead of the default navigate away.

I can open up a PR upstream but there may be another way we could do this, with just an app wide gesture listener on our side doing the same thing basically. I'm not super familiar with the expo + tv side of things but

lancechant: I see, yeah there is a global way it works already, and the react-native-tvos sort of handles that if I'm correct, if anything you got to just hook it, and maybe ensure it's called in every page it's needed, I noticed that a lot of screens don't have it, but I think you need to add the logic for it to be handled, I think it's about 10 lines or so.

Just thinking a patch like this is going to have to be maintained every update, and is possibly a manual process to fix if anything goes wrong, or renamed (Seen it happen on another patch, hence the worry)

stevebyatt10: Alright i've gone back through and found the actual issue, jumped the gun and went straight to native code 😅.

Heavily simplified the logic for back button handling now since there was some overlap.

Thanks for calling out the patch

lancechant: Lol all good, happy to have ensured the correct work was done

components/video-player/controls/hooks/useRemoteControl.ts
```diff
@@ -106,61 +93,38 @@ export function useRemoteControl({
...
   }, []);
 
   // TV remote control handling (no-op on non-TV platforms)
   useTVEventHandler((evt) => {
     if (!evt) return;
 
-    // Back/menu is handled by BackHandler above, but keep this for tvOS menu button
+    // Back/menu is handled by useTVBackPress above. Keep this handler focused
+    // on remote-control events like play/pause, D-pad, and long seek.
     if (evt.eventType === "menu") {
```
lancechant, line 127: So does menu not have any actions now? what button is the menu, lol?

lancechant (approved):

lancechant:
Might need you to align the back handling on here #1557 @stevebyatt10

components/video-player/controls/hooks/useRemoteControl.ts
stevebyatt10: image
Lol a bit confusing, but menu is basically back on apple tv. So it comes through the same back handler, removed from here so to not worry about duplicate event handling

stevebyatt10:
@lancechant  Yep i can rebase that PR onto this one or once its merged and make sure its all good 🙏

components/video-player/controls/hooks/useRemoteControl.ts
lancechant: The image ain't loading 🤣 but I get what you saying I think. So it's ignored here so it doesn't conflict with the back handler correct?

lancechant:
Hey @stevebyatt10 Okay cool, i've merged. not sure which PR you want in first, but just a reminder to update the back handling from my pr😁

stevebyatt10:
@lancechant awesome! I think best order is this back handler, the focus fix on player and then top shelf. 

With this one in I can rebase your QR code PR too and test on Apple TV too 🙏

lancechant:
Perfect, let me know when ready

stevebyatt10:
Yep all are good to go 🙏 @lancechant

lancechant:
Cool done, let me know how the rebase etc goes 😎

stevebyatt10:
@lancechant It seems like these changes alone don't actually fix it on apple tv.
It was because i had the patch still applied in node_modules that worked 😅 

After I rebased and tested the back logic, it wasn't being handled and just navigating away again.

https://github.com/react-native-tvos/react-native-tvos/issues/988#issuecomment-3336219366
This seems like it does require a patch to work on tvos and `react-native-screens`

lancechant:
@stevebyatt10 Ahhh dangit those node modules, alright at least it is a known thing, wish it was addressed as that is oldish by now

lancechant:
Wait @stevebyatt10 Can you try a newer version of react-native-screens? it might be addressed there before you do the patch again?

lancechant:
Also maybe try a newer version of `react-tvos` it looks like there was some new logic to handle menu buttons from what I see @stevebyatt10

stevebyatt10:
Yeah I did have a look at newer version of `react-native-screens` and some of the commits but they don't look like they address any of this unfortunately. 

Also took a look at the latest rn tvos version `0.85.3-0`, the main issue is really because of react-native-screens’ `RNSNavigationController` it doesn't look like any of the new changes would help us hear unfortunately

lancechant:
Dang okay, I thought it might have been addressed as say better back handling for IOS, but I guess the patch for now

PR #1623, opened 2026-05-31
fix(item): dedupe top people sections by id

Gauvino:
## Problem

The item detail screen renders a "More with <actor>" section for each of the top 3 people. Jellyfin can list the same person more than once (e.g. credited as both actor and director/writer), so the same person could appear twice — producing a duplicate section and a React "two children with the same key" warning.

## Fix

Dedupe the people list by `Id` before taking the top 3, so we always surface 3 *distinct* people.

No behavior change otherwise.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed duplicate people entries appearing in sections; now displays only unique people up to a maximum of three.


Alexk2309 (approved):

Gauvino:
@coderabbitai full review

PR #1586, opened 2026-05-22
feat(player): add chapter markers and chapter list

Gauvino:
# 📦 Pull Request

## 🔖 Summary
Surface Jellyfin chapter markers in the native video player — tick marks on the seek bar, a chapter list sheet to jump between chapters, the active chapter name in the controls header, and the chapter name on the trickplay bubble while scrubbing. Works online and offline.

## 🏷️ Ticket / Issue
- Supersedes #1299
- Closes #34 (Chapters in videos — original request)

## 🛠️ What's Changed
- Type: `feat`
- Scope: `player`
- Adds chapter tick marks on the progress bar, a chapter list sheet, a live chapter-name label in the player header, and the chapter name on the trickplay bubble while scrubbing. Refreshes the cached `BaseItemDto.Chapters` at download time so offline playback gets the same UI.

## 📋 Details

Jellyfin exposes chapter markers on items, but the native player never surfaced them. This PR wires the data end-to-end with a strong split between pure helpers, presentational components, and integration glue.

### Architecture (4 commits)

| Commit | Layer | Files |
|---|---|---|
| `8d33aa3b` | Pure helpers + unit tests | `utils/chapters.ts`, `utils/chapters.test.ts` |
| `5f64ce49` | Presentational components (player-agnostic) | `components/chapters/ChapterTicks.tsx`, `components/chapters/ChapterList.tsx` |
| `98b90f5b` | Integration into the native player | `components/video-player/controls/{BottomControls,Controls,TrickplayBubble}.tsx`, `components/video-player/controls/hooks/useVideoSlider.ts`, `translations/en.json` |
| `558fb418` | Offline robustness | `components/DownloadItem.tsx` |

### Helpers (`utils/chapters.ts`)

Dependency-free, fully tested under `bun test`:
- `chapterMarkers(chapters, durationMs)` — markers within range, with precomputed `percent` for slider overlays.
- `chapterStartsMs(chapters)` — sorted start positions in ms (skips entries without `StartPositionTicks`).
- `currentChapterIndex(positionMs, chapters)` — active chapter index (-1 before the first chapter).
- `chapterNameAt(positionMs, chapters)` — active chapter name, or `null` if missing/unnamed.
- `sortedChapters(chapters)` — chapter entries paired with ms start.
- `formatChapterTime(positionMs)` — `m:ss` / `h:mm:ss` label.

All tolerate `null` / `undefined` / `[]`. **17 unit tests** cover sort order, boundary positions, missing fields and out-of-range inputs.

[...]

PR #1624, opened 2026-05-31
fix(chapters): keep landscape when opening chapter list on iOS

Gauvino:
## Problem

On iOS, opening the chapter list while the player is in landscape rotates the whole app back to portrait. The list then shows in portrait instead of over the landscape video.

## Cause

`components/chapters/ChapterList.tsx` renders a React Native `<Modal>` without `supportedOrientations`. iOS defaults modals to portrait-only, so presenting one in landscape forces the app to rotate.

## Fix

Set `supportedOrientations={["portrait", "landscape"]}` on the modal so it opens in the current orientation. The prop is iOS-only; Android ignores it, so no cross-platform change.

Follow-up to the merged chapters feature (#1586).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed modal to remain usable when device orientation changes between portrait and landscape.

### DIFF
--- a/components/chapters/ChapterList.tsx
+++ b/components/chapters/ChapterList.tsx
@@ -74,6 +74,9 @@ function ChapterListComponent({
       transparent
       animationType='slide'
       onRequestClose={onClose}
+      // iOS defaults <Modal> to portrait-only; without this it rotates the app
+      // back to portrait when opened from the landscape player. Android ignores it.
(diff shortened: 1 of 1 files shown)


Alexk2309 (approved):

Gauvino:
@coderabbitai full review

PR #1640, opened 2026-06-01
fix: remove empty sections from section tree

lostb1t:
Empty sections would still take up space. This PR hides the section that are empty. 

Because loading logic lives in the section itself we have to signal the wrapper for it. (i dont wanna refactor)

before:
IMG_1799

after:

image

### DIFF
--- a/components/home/Home.tsx
+++ b/components/home/Home.tsx
@@ -490,12 +490,18 @@ const HomeMobile = () => {
 
   const sections = settings?.home?.sections ? customSections : defaultSections;
 
+  const serializeSectionKey = useCallback(
+    (queryKey: (string | undefined | null)[]) =>
+      JSON.stringify(queryKey.map((k) => k ?? null)),
+    [],
+  );
+
   // Get all high priority section keys and check if all have loaded
   const highPrioritySectionKeys = useMemo(() => {
     return sections
       .filter((s) => s.priority === 1)
-      .map((s) => s.queryKey.join("-"));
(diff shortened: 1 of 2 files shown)


fredrikburmester:
Could you follow the PR template please.

components/home/Home.tsx
```diff
@@ -675,15 +705,19 @@ const HomeMobile = () => {
                       ? () => markSectionLoaded(section.queryKey)
                       : undefined
                   }
+                  onEmptyChange={(isEmpty) =>
+                    handleSectionEmptyChange(section.queryKey, isEmpty)
```
fredrikburmester, line 709: If this sets isEmpty the below streamystats section will also be hidden. Is that intended? Maybe we should lift the streamystats section out of the `<View />` or put the isHidden on the `<InfiniteScrollingCollectionList />`?

fredrikburmester (changes requested):

components/home/Home.tsx
lostb1t: so streamystatssection is just another section but from a different source? dont have streamystats so havent seen it before.

If its basicly just another section then i suggest we lift the inlining and just add it to the sections list itself?

if you want to keep it inline (to group by row title i guess?) maybe still lift it and group by some key.

il see if i can setup streamystats so i can actually see how this is rendered

lostb1t:
closing this. Too much refactor for an edge case

PR #1649, opened 2026-06-01
fix(player): correct prev/next episode buttons at boundaries & for missing episodes

Gauvino:
# 📦 Pull Request

[AI Assisted](#)

## 📝 Description

The player's previous/next episode buttons were derived from the **length** of the `getEpisodes({ adjacentTo })` response (`hooks/usePlaybackManager.ts`). That heuristic breaks at series boundaries: `adjacentTo` does not guarantee a fixed `[prev, current, next]` shape, so on the **first episode** (e.g. S1E1) the current episode itself was surfaced as "previous" and a bogus **previous** button appeared. It also offered **"Virtual" (missing) episodes** — like an empty Specials (Season 0) entry with no media file — as navigation targets, and `nextItem` could make **autoplay** advance into a missing episode.

Fix: derive prev/next from the **current item's real index** in the adjacent list and skip any neighbour that is the current item or has `LocationType === "Virtual"`.

Covered edge cases:
- First episode → no false "previous" button.
- Last episode → no false "next" button.
- Missing / empty Specials → never offered as prev/next.
- Autoplay no longer advances into a missing episode.
- Offline path benefits too (now index-based for all list shapes).

## 🏷️ Ticket / Issue

N/A — bug found during testing. Previous button wrongly shown on Black Clover S1E1 (Specials/Season 0 empty).

Without (button previously showing and throwing falied to get sream url) : 
qemu-system-x86_64_kooYScNAUK
With (not showing on first episode) : 
qemu-system-x86_64_kOG0S377Qk

## ✅ Checklist

- [ ] Verified that changes behave as expected for all platforms _(verifying live)_
- [x] I've declared if AI was used to assist with this PR (AI-Assisted badge above)

## 🔍 Testing Instructions

1. Open a series whose **first** episode has no real previous (e.g. one with an empty/absent Specials season). Play **S1E1**.
2. Verify **no** "previous episode" button is shown in the player header.
3. Play a middle episode → both previous and next buttons appear and navigate correctly.
4. Play the **last** episode of the last season → **no** "next" button.
5. If the series has a missing episode (Jellyfin "Virtual" placeholder) adjacent to the current one, verify it is **not** offered as prev/next and autoplay does not jump to it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced episode navigation to more accurately determine the current episode's position within available episodes and identify valid neighboring episodes for previous/next selection. The system now properly validates that previous and next episode navigation options exist and are selectable before presenting them as choices, while excluding certain unavailable episode types from the navigation flow.

### DIFF
--- a/hooks/usePlaybackManager.ts
+++ b/hooks/usePlaybackManager.ts
@@ -109,30 +109,35 @@ export const usePlaybackManager = ({
     staleTime: 0,
   });
 
+  /**
+   * Derive prev/next from the current item's real position in the adjacent
+   * list rather than from the array length. `getEpisodes({ adjacentTo })` does
+   * not guarantee a fixed [prev, current, next] shape — at the first/last
+   * episode it can still return the current item as the first/last entry — so
+   * length-based indexing wrongly surfaces the current episode as "previous".
+   */
+  const currentIndex = useMemo(
+    () => adjacentItems?.findIndex((e) => e.Id === item?.Id) ?? -1,
+    [adjacentItems, item],
+  );
+
+  /** A neighbour is only navigable if it has an actual media file (not a
+   * "Virtual"/missing episode placeholder, e.g. an absent Special). */
+  const isNavigable = (episode?: BaseItemDto | null): episode is BaseItemDto =>
+    !!episode && episode.Id !== item?.Id && episode.LocationType !== "Virtual";
+
   const previousItem = useMemo(() => {
-    if (!adjacentItems || adjacentItems.length <= 1) {
-      return null;
-    }
-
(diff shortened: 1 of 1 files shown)


Gauvino:
@coderabbitai full review

lancechant (approved):

Alexk2309:
One thing to be careful of here is that we also use this logic for offline series. We may need to replicate this or ensure the new functionality works with it.

Gauvino:
> One thing to be careful of here is that we also use this logic for offline series. We may need to replicate this or ensure the new functionality works with it.

I have tested on android offline and it work, it does apply to it if im not that bad

PR #1650, opened 2026-06-02
fix(platform-dropdown-component): Auto-size iOS trigger wrapper to pr…

Alexk2309:
fix(dropdown): auto-size iOS trigger wrapper to prevent stale label wrapping

# 📦 Pull Request

[AI Assisted](#)

## 📝 Description

Fixes a layout bug in the season dropdown on iOS where the label wrapped onto two lines:

```
Season
1
```

instead of `Season 1`.

**Root cause:** the previous iOS `PlatformDropdown` implementation pinned the wrapper to a `useState`-stored measurement of the trigger taken via `onLayout`. On `SeasonDropdown`'s first render, `seasonIndex` is `undefined`, so the trigger reads `"Season "` (no number). The wrapper gets pinned to that narrow width. A `useEffect` then selects the first season, the trigger re-renders as `"Season 1"`, but the wrapper (and therefore the SwiftUI `<Menu>` label frame inside `<Host>`) is already locked at the stale width — so the new text wraps. This affects any dropdown whose label content can change after mount (bitrate, audio track, subtitle selectors, etc.).

**Fix:** replace the measured-pin approach with an in-flow invisible copy of the trigger that sizes the wrapper naturally on every render. The `<Host>` still overlays the real interactive `<Menu>` via `absoluteFill`, so:

- The wrapper auto-sizes to current trigger content — no state, no race, no stale measurement.
- The original SDK 55 fix (preventing stacked dropdowns from collapsing to 0 in the download sheet) is preserved: the wrapper still has a concrete non-zero size, and `<Host>`'s `absoluteFill` still breaks the flex circular dependency.
- Removed now-dead code: `triggerSize` state, `handleMeasureTrigger`, `useState` and `LayoutChangeEvent` imports.

## 🏷️ Ticket / Issue

N/A

## ✅ Checklist

## 🔍 Testing Instructions

**Regression (the bug this PR fixes):**

1. Open the app on iOS and navigate to any series with multiple seasons.
2. Confirm the season picker shows `Season 1` (or whichever season was last selected) on a single line — not `Season` over `1`.
3. Open the dropdown and switch seasons; the label updates and continues to render on a single line.

**Preserve previous fix (stacked dropdowns):**

1. From a series episode, open the download sheet (download icon → choose a download quality flow).
2. Confirm **all** stacked dropdowns are visible and tappable: Quality, Video, Audio, Subtitle. None should collapse / disappear.

**Other dropdown call sites — smoke test:**

1. Library filter dropdowns on the libraries tab.
2. Audio / subtitle / bitrate selectors during playback.
3. Jellyse

## Summary by CodeRabbit

* **Refactor**
  * Simplified dropdown layout calculation logic for iOS devices while maintaining existing functionality across all platforms.

### DIFF
--- a/components/PlatformDropdown.tsx
+++ b/components/PlatformDropdown.tsx
@@ -1,13 +1,7 @@
 import { Ionicons } from "@expo/vector-icons";
 import { BottomSheetScrollView } from "@gorhom/bottom-sheet";
-import React, { useEffect, useState } from "react";
-import {
-  type LayoutChangeEvent,
-  Platform,
-  StyleSheet,
-  TouchableOpacity,
-  View,
-} from "react-native";
+import React, { useEffect } from "react";
+import { Platform, StyleSheet, TouchableOpacity, View } from "react-native";
 import { useSafeAreaInsets } from "react-native-safe-area-context";
 import { Text } from "@/components/common/Text";
 import { useGlobalModal } from "@/providers/GlobalModalProvider";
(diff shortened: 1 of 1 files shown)


Gauvino:
@coderabbitai full review

Gauvino:
@Alexk2309 when making pr that touch ui could you please provide screenshot with before and after, but otherwise looks good

Gauvino (approved):

PR #1652, opened 2026-06-02
fix(downloads): poster images not showing for downloaded content

Alexk2309:
# 📦 Pull Request

[AI Assisted](#)

## 📝 Description

Downloaded items were rendering the placeholder poster instead of the actual cover art. Root cause is in `hooks/useImageStorage.ts`: the `fetch(url) → response.blob() → FileReader.readAsDataURL(blob)` pipeline silently resolves to an empty payload under React Native's New Architecture, so we end up writing an empty/invalid base64 string into MMKV.

This PR swaps just the download-and-decode step for `expo-file-system`'s `File.downloadFileAsync(url) → file.base64()` (the same primitive already used for trickplay and subtitles), and cleans up the temp file in `Paths.cache` afterward. The MMKV key/value format is unchanged (still base64 stored under the item id), so every consumer (`MovieCard`, `SeriesCard`, `DownloadCard`, offline series page) keeps working without modification.

Note: storing posters as base64 in MMKV is not ideal long-term (memory-mapped, ~33% size overhead) — but that's a separate refactor. This PR is the minimal fix to restore the broken behavior.

## 🏷️ Ticket / Issue

N/A

## ✅ Checklist

## 🔍 Testing Instructions

1. Start from a clean state (or clear the existing downloaded image cache by deleting and re-downloading an item).
2. Download a movie and a TV episode from a Jellyfin library.
3. Open the Downloads tab and verify the movie poster, series poster, and per-episode card all show the real artwork instead of the placeholder.
4. Open the downloaded series detail page while offline and verify the backdrop/poster renders.
5. Force-quit and relaunch the app; verify posters still render (i.e., persisted in MMKV correctly).
6. Repeat on both iOS and Android with the New Architecture enabled.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling for image downloads and caching; failed downloads now handled gracefully with appropriate fallback behavior.

* **Refactor**
  * Optimized image storage mechanism for improved performance and more efficient file management.

### DIFF
--- a/hooks/useImageStorage.ts
+++ b/hooks/useImageStorage.ts
@@ -1,3 +1,4 @@
+import { File, Paths } from "expo-file-system";
 import { useCallback } from "react";
 import { storage } from "@/utils/mmkv";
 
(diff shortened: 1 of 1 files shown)


Gauvino (approved):

Gauvino:
@coderabbitai full review

PR #1653, opened 2026-06-02
fix(player): respect safe area in chapter list and refine UX

Alexk2309:
# 📦 Pull Request

[AI Assisted](#)

## 📝 Description

Polish for the in-player chapter list sheet and the chapter toggle button in the bottom controls.

- Inset the chapter list sheet by the left/right/bottom safe-area so text and the close button no longer slide under the notch / Dynamic Island / home indicator in landscape.
- Remove the dimmed backdrop above the sheet so the video stays fully visible behind it (the sheet still dismisses on backdrop tap).
- Reorder the bottom controls so the chapter icon sits **after** Skip Intro / Skip Credits / Next Episode instead of in front of them.
- Add pressed-opacity feedback on the chapter icon so it visibly responds to taps.
- Drop the unused `chapterPositions` prop from `BottomControls` (and its pass-through in `Controls.tsx`) — the component derives markers from `chapters` via `chapterMarkers`.

## 🏷️ Ticket / Issue

N/A

Before: 

Note that the Dynamic Island isn’t visible here, but it does obstruct this on a real device.

IMG_9584

After

IMG_9583

## ✅ Checklist

- [ ] Code passes lint/formatting and type checks (`tsc`/`biome`)
- [ ] No secrets, hardcoded credentials, or private config files are included

## 🔍 Testing Instructions

1. Start playback of any item that has more than one chapter.
2. Rotate to landscape on a notched iPhone (or any device with non-zero side insets).
3. Open the chapter list via the bookmarks icon in the bottom controls.
   - Verify the sheet's left/right edges sit inside the safe area (not under the notch).
   - Verify the bottom row clears the home indicator.
   - Verify the area above the sheet is **not** dimmed — the video remains visible.
   - Verify tapping above the sheet still dismisses it.
4. While the bottom controls are visible, confirm the chapter (bookmarks) icon appears to the **right** of Skip Intro / Skip Credits / Next Episode buttons.
5. Press and hold the chapter icon — it should fade to a lower opacity while held and return to full opacity on release.
6. Open an episode where Skip Intro / Skip Credits trigger and confirm those buttons still appear and behave the same as before.
7. Run `bun run typecheck` and `bun run check` — both should pass.

### DIFF
--- a/components/video-player/controls/BottomControls.tsx
+++ b/components/video-player/controls/BottomControls.tsx
@@ -75,9 +75,6 @@ interface BottomControlsProps {
     minutes: number;
     seconds: number;
   };
-
-  // Chapter props
-  chapterPositions?: number[];
 }
 
 export const BottomControls: FC<BottomControlsProps> = ({
@@ -111,7 +108,6 @@ export const BottomControls: FC<BottomControlsProps> = ({
   trickPlayUrl,
   trickplayInfo,
   time,
-  chapterPositions = [],
 }) => {
   const { settings } = useSettings();
   const { t } = useTranslation();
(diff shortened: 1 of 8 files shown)


components/video-player/controls/BottomControls.tsx
```diff
@@ -75,9 +75,6 @@ interface BottomControlsProps {
     minutes: number;
     seconds: number;
   };
-
-  // Chapter props
-  chapterPositions?: number[];
```
Alexk2309, line 80: Unused

Gauvino:
My bad didn't notice the bug

Gauvino (approved):

PR #1658, opened 2026-06-02
fix(downloads): follow-ups to #1547 review (stale offline list + dead code)

Gauvino:
> [!WARNING]
> **Draft / not for merge as-is.** This branch is stacked on top of #1547 (`Kyliroco:perf/download-database`), which is a cross-fork PR. Because GitHub diffs against `develop`, the diff below also shows all of #1547's commits. The **only new work here is the 2 commits listed below** — opened mainly to exercise CodeRabbit pro-plus and to validate the follow-up fixes before they land on #1547.

## New commits (the actual change)

- `b05d684e` **fix(downloads): read live DB in offline episode queries to avoid stale list**
- `602da2b8` **chore(downloads): remove unused _invalidateCaches helper**

## Context — addresses the open review comments on #1547

**1. Stale offline episode/season list (Copilot, ×4 on `EpisodeList` + `SeasonEpisodesCarousel`).**
#1547's perf refactor switched the offline query reads from `getDownloadedItems()` (live, cached DB) to the `DownloadProvider.downloadedItems` snapshot, which only refreshes on the provider `refreshKey`. But `usePlaybackManager` saves offline progress/played/unplayed via `updateDownloadedItem()` and then `invalidateQueries(["episodes"])` **without** calling `triggerRefresh()`. So the refetch re-read a stale snapshot and the list never reflected the change — silently breaking the PR's own "refetch from updated local database" intent. Fix: read `getAllDownloadedItems()` (cached → cheap; rebuilds only after a DB write) inside the queryFn, restoring correctness while keeping the perf win.

**2. `_invalidateCaches()` unused (Copilot, `database.ts`).** Removed — `saveDownloadsDatabase` already invalidates the derived caches inline; the helper had zero call sites repo-wide.

## On the CodeRabbit `startPosition` concern (`direct-player.tsx`) — verified NOT a bug, no change

The JS one-shot seek (`initialSeekDoneRef`, `tracksReady`) does not re-seek on episode switch via `setParams`. **But resume is applied by the native layer**: every source change fires Expo `Prop("source")` → `loadVideo` → `renderer.load`, which sets MPV `start` then `loadfile … replace` on both Android (`MPVLayerRenderer.kt`) and iOS (`MPVLayerRenderer.swift`). The new episode starts at its correct position. Minor edge: native `loadVideo` early-returns if the URL is unchanged, so re-selecting the currently-playing episode won't re-seek — not the reported flow.

## Release Notes

* **Bug Fixes**
  * Fixed race condition that could cause stale item data to persist when switching between items
  * Improved streaming stability when navigating between content

* **Performance Improvements**
  * Optimized downloaded items lookup performance through improved caching mechanisms
  * Enhanced offline mode data handling for faster access to cached downloads


Gauvino:
@coderabbitai full review

Gauvino:
Superseded — these commits were pushed directly onto #1547 (`perf/download-database`). This stacked test PR is no longer needed.

PR #1666, opened 2026-06-04
fix(downloads): Use mediaSource.Id instead of item.Id in direct download URL

boolemancer:
# 📦 Pull Request

[AI Assisted](#)
I used Github Copilot to familiarize myself with the structure of the codebase, but wrote and validated the changes myself, and no AI assistance was used when writing this PR description.

## 📝 Description

Fix a bug where the version selection is ignored for direct downloads (i.e. the Quality is set to 'Max').

For some background, I just came across this app recently when trying out different Jellyfin clients and was happy to find one that finally seemed to have some good support for downloading things. I was especially pleased to find out that it had support for selecting which version of a video to download, since that's something I've been looking for.

Unfortunately, when I went to actually try it out, I noticed that it always seemed to download the same file, regardless of which version (mediaSource) I selected. Rather than just open a bug and hope someone cares enough to look into it eventually, I figured I'd take a look myself and spent the evening getting the dev environment set up to test it out.

Technical details: In the `getDownloadUrl` implementation, a `mediaSource` is passed in to specify which 'version' of the item to download, but in the `Max` quality scenario, the `mediaSource` is entirely ignored when building the URL which ultimately gets downloaded. Fortunately, the `mediaSource` IDs are actually also BaseItem IDs that represent the underlying files that get grouped together to form the one item with different versions, and those IDs should also be valid in the same `Items` API.

With that, the fix is simple, just replace the `item.Id` with `mediaSource.Id`, and you direct-download the version the user selected instead of always downloading the same one .

Validation was done on a physical Android phone against a Jellyfin server running version 10.11.10.

## 🏷️ Ticket / Issue

I did not find an existing issue for the bug, but I can create one if necessary.

N/A, no UX changes

## ✅ Checklist

> - [ ] Verified that changes behave as expected for all platforms

I only have access to an Android device, so I have not validated the changes against any other platforms, but I did validate the changes on a physical Android device. That said, I do not believe that this change should have any platform-specific behavior.

## 🔍 Testing Instructions

1. Open up an item in your library which has multiple versions (e.g. 720p and 4K)
2. Select a version to download at Max quality
3. Press the download button
4. Go to the downloads page and verify that the download file size matches the expected size for the version of the item you selected
5. (optional) Delete/cancel that download and repeat the steps with a different version to verify that that the download file size is different

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue with the download URL generation that was preventing downloads from functioning correctly.


lancechant (approved):

lancechant:
Hey @boolemancer Can you just show somehow the difference? and I assume only max was affected? the others were fine as it's just passed in to the download stream url?

boolemancer:
Here's some screen recordings showing the repro and the fix, following the testing instructions above. The 1080p version should be ~15 GiB and the 720p version should be ~2.3GiB

Repro of the bug in the release build of Streamyfin:

https://github.com/user-attachments/assets/0391be29-e36c-4021-9b2f-cb2d64ce379a


https://github.com/user-attachments/assets/7ecceb1c-242a-4d59-ae93-f19ea87625f7

Note that the download size is the same for both versions.

Same steps with the fix from this PR:

https://github.com/user-attachments/assets/6ef86013-0205-4a7f-9539-85268a99ffa6


https://github.com/user-attachments/assets/62dbd639-b77d-47c0-b5e8-608d077edaa5

Note that the download size is different between the two versions.

> and I assume only max was affected? the others were fine as it's just passed in to the download stream url?

Yeah, the others seem fine. I just double checked by looking at the transcoding logs on the server and it looks like it's transcoding the correct files based on the selected version.

lancechant (approved):

Gauvino (approved):

Gauvino:
@coderabbitai full review

PR #1659, opened 2026-06-02
fix: pre-release bug-fix batch (filters, Quick Connect, auth, downloads, i18n)

Gauvino:
# 📦 Pull Request

[AI Assisted](#)

## 📝 Description

Pre-release bug-fix batch — one commit per fix (15 commits):

| Fix | Summary |
|---|---|
| **Filters never opened** (new arch) | `BottomSheetModal.present()` from a `useEffect` silently no-ops on Fabric + Reanimated 4. Filter sheets now present imperatively from the press handler, open instantly with a loader; same fix for the bitrate/source/track sheets. |
| **Filter search ate keystrokes** | On large option lists (2000+ tags) each keystroke re-filtered + re-rendered the list, desyncing the controlled input (lost/reappearing letters). `useDeferredValue` + memoized rows. |
| **Quick Connect code stuck on screen** | The code lived in a native Alert (no programmatic dismiss). Now an in-app sheet that auto-closes once the session is authorized, with tap-to-copy (expo-clipboard). Polling also stops when leaving the login screen or going back to server selection. |
| **Save account asked before the login** | The protection picker showed before the credentials were even checked (and Quick Connect couldn't save at all). Login flows now flag the intent; a global modal asks AFTER the session is authorized — identical for password and Quick Connect. Cancelling saves nothing. |
| **Session not cleared on 401** | A response interceptor clears the session once when the server revokes the token — no more 401 spam / half-authenticated state. Saved accounts survive. |
| **Offline splash hang (75-120 s)** | Startup no longer awaits `getCurrentUser` (axios has no timeout); cached user renders immediately, validation runs in the background. |
| **Admin-locked settings changeable** | `updateSettings` strips plugin-locked keys at write time; the Streamystats screen shows the live admin value and disables locked controls. |
| **Delete all downloads: no confirmation** | Destructive action now asks first. |
| **Storage bar showed 0.00%** | Size computed from live files on disk, auto-refreshes every 10 s, refreshes right after delete-all, and in-progress downloads count as app usage. |
| **Recycled series poster** | Poster cache read keyed on `SeriesId` (was frozen on first render). |
| **Duplicate screens from rapid taps** | Focus-based push guard in `useAppRouter` (no timers). |
| **Max-update-depth via filter options** | `useFilterOptions` memoized + debug logs removed. |
| **Deprecated notification flags** | `shouldShowAlert` → `shouldShowBanner`/`shouldShowList`; foreground listener no longer dumps full payloads (deprecated `dataString` getter). |
[...]

PR #1669, opened 2026-06-04
feat(settings): unified server URL resolver + shared field

Gauvino:
[AI Assisted](#)

# 📦 Pull Request

## 📝 Description
Unify how server URLs are entered across the app. Today each integration parses the URL its own way — Jellyfin uses the SDK discovery, Jellyseerr probes `/api/v1/status` on the raw string, Streamystats only strips a trailing slash — so the user has to know the exact protocol/port. This adds a generic resolver + a shared field so typing `media.example.com` (or `https://media.example.com`, `host:port`, …) "just works".

**Core (`utils/serverUrl/`)**
- `getServerUrlCandidates` — loose input → ordered candidates: https first, http fallback, explicit port/path preserved, **no Jellyfin-specific ports** (service-agnostic, unlike the SDK's `getAddressCandidates`).
- `resolveServerUrl(input, probe)` — probes candidates in parallel with a per-candidate timeout; returns the first that validates, else the most actionable failure (`version-too-low` › `wrong-service` › `unreachable`).
- `isVersionBelow` — numeric semver compare (fixes the `"2.10.0" < "2.0.0"` string-compare bug).
- `jellyseerrProbe` — `/api/v1/status`, requires ≥ 2.0.0.

**Hook + UI**
- `useServerUrlResolver(probe)` — `idle → resolving → ok | error` state machine with cancellation.
- `ServerUrlField` — shared input: auto-resolves on blur, inline status chip (tap to re-test), shows the resolved canonical URL, and persists the **resolved** URL (not the raw input).

**Adoption**
- Jellyseerr settings now use `ServerUrlField`; login uses the resolved URL. Addresses the "isn't this done elsewhere already?" duplication raised on #1659.

Follow-ups (same probe contract → drop-in): Streamystats, the Jellyfin login screen, MerlinSearch.

## 🏷️ Ticket / Issue
Related: #1659 (review discussion on unifying server-URL handling)

## ✅ Checklist
- [ ] Verified on device (draft — pending)

## 🔍 Testing Instructions
1. Settings → Plugins → Jellyseerr.
2. Type just the host (e.g. `media.example.com`) and leave the field → resolves to `https://…`, ✓ chip, canonical URL shown below.
3. Type a bad host → ✗ "unreachable"; tap the chip to retry.
4. Point at a host that responds but isn't jellyseerr → "not the expected server".
5. Log in → succeeds against the resolved URL; reopen settings → URL persisted.

## 📝 Additional Notes
Draft: core + Jellyseerr only. Streamystats / Jellyfin login / Merlin to follow once the field is validated on device. New i18n keys added to `en.json` only (other locales fall back to English via i18next; Crowdin sync later).

### DIFF
--- a/components/common/ServerUrlField.tsx
+++ b/components/common/ServerUrlField.tsx
@@ -1,0 +1,177 @@
+import { Ionicons } from "@expo/vector-icons";
+import { useCallback, useRef } from "react";
+import { useTranslation } from "react-i18next";
+import { ActivityIndicator, Pressable, View } from "react-native";
+import {
+  type ServerUrlResolverState,
+  useServerUrlResolver,
+} from "@/hooks/useServerUrlResolver";
+import type {
+  ResolveFailureReason,
+  ResolveOptions,
+} from "@/utils/serverUrl/resolve";
+import type { ServerProbe } from "@/utils/serverUrl/types";
+import { Input } from "./Input";
+import { Text } from "./Text";
+
+interface ServerUrlFieldProps {
+  /** Raw user input (controlled). */
+  value: string;
+  onChangeText: (text: string) => void;
+  /** Service-specific validator. Pass a stable (module-level) reference. */
+  probe: ServerProbe;
+  /** Called with the canonical URL once a candidate validates. */
+  onResolved?: (url: string, meta?: Record<string, unknown>) => void;
+  label?: string;
+  hint?: string;
+  placeholder?: string;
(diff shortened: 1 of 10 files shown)


Gauvino:
Superseded by https://github.com/streamyfin/streamyfin/pull/1670 — moved from my fork to an org branch (`feat/unified-server-url` on streamyfin/streamyfin). Same content.

PR #1678, opened 2026-06-05
fix(series): filter Next Up items by SeriesId client-side

Alexk2309:
# 📦 Pull Request

[AI Assisted](#)

## 📝 Description

The mobile series detail page's "Next Up" section calls `GET /Shows/NextUp` with a `seriesId` filter, but in practice the response can contain episodes from other series — the same content as the home tab's Next Up row. The result is that opening a series (e.g. from search) shows a "Next Up" row of unrelated episodes.

This PR adds a defensive client-side filter on the response so we only ever show episodes whose `SeriesId` matches the current page's `seriesId`. When nothing matches (e.g. an unwatched series, or every returned item belongs to a different show), the section is hidden entirely instead of rendering an empty "No items to display" block.

The filter is a no-op when the server's response is already correctly scoped.

## 🏷️ Ticket / Issue

N/A

N/A — section visibility only, no visual changes when populated.

## ✅ Checklist

## 🔍 Testing Instructions

1. From the Search tab, search for a series you have **never watched** and tap it.
2. On the series detail page, scroll to the area above the season picker.
3. **Before this fix:** a "Next Up" row appears showing episodes from other series (matching what's on the Home tab's Next Up row).
4. **After this fix:** no "Next Up" section is rendered for an unwatched series.
5. Navigate to a series you **are** currently watching (e.g. from Home → Continue Watching).
6. Verify the "Next Up" section appears and that every episode in it belongs to the currently-opened series.
7. Repeat from the Libraries, Favorites, and Watchlists tabs to confirm parity.
8. Confirm TV (no changes) and offline mode (`NextUp` not rendered) still behave the same.

### DIFF
--- a/components/series/NextUp.tsx
+++ b/components/series/NextUp.tsx
@@ -3,6 +3,7 @@ import { FlashList } from "@shopify/flash-list";
 import { useQuery } from "@tanstack/react-query";
 import { useAtom } from "jotai";
 import type React from "react";
+import { useMemo } from "react";
 import { useTranslation } from "react-i18next";
 import { View } from "react-native";
 import { apiAtom, userAtom } from "@/providers/JellyfinProvider";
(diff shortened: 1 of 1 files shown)


Simon-Eklundh:
I'm not against this, but I can't seem to replicate the issue. 

could already be fixed, or could be a problem on a specific server version I'm not on. what version have you seen this on?

PR #1719, opened 2026-06-11
chore(deps): upgrade TypeScript to 6.0.3 and align Expo SDK 56 packages

Gauvino:
# 📦 Pull Request

[AI Assisted](#)

> [!NOTE]
> This PR also contains the Expo SDK 56 dependency alignment from #1720, which was merged into this branch.

## 📝 Description

Upgrades TypeScript from 5.9.3 to 6.0.3 (the version expo doctor expects for SDK 56) and aligns all dependency versions with the installed Expo SDK 56, bringing `expo doctor` from 3 failing checks down to 1. No application code changes and no Expo SDK jump — all dependency bumps are patch-level within SDK 56.

### TypeScript 6 upgrade

- **`typescript` 5.9.3 → 6.0.3** (devDependency)
- **`@types/node@^18.19.130` added as a direct devDependency** — it was previously only available as a transitively hoisted package (same version), which TS 5.9 picked up via automatic `@types` inclusion
- **`tsconfig.json`: explicit `"types": ["node", "jest"]`** — TS 6.0 changed the default of `compilerOptions.types` from "all packages under `node_modules/@types`" to `[]` (see the TS 6.0 release notes). Without it, globals from `@types/node` (`Buffer`, `NodeJS.*`, `process.exit` typed as never-returning) and `@types/jest` (`it`, `expect`) no longer resolve, producing 20 typecheck errors
- **`scripts/typecheck.js`: force `--pretty false`** — TS 6.0 enables pretty diagnostics even when output is piped, which broke the script's line-based error parser and garbled its output

### Expo SDK 56 alignment (from #1720)

- **18 packages patch-aligned** via `bunx expo install --fix` (`@expo/metro-runtime`, `@expo/ui`, `expo`, `expo-asset`, `expo-audio`, `expo-background-task`, `expo-build-properties`, `expo-camera`, `expo-constants`, `expo-dev-client`, `expo-font`, `expo-image`, `expo-linking`, `expo-location`, `expo-notifications`, `expo-router`, `expo-sharing`, `expo-task-manager`). `expo.install.exclude` (react-native, react-native-screens) untouched and respected.
- **app.json: added `expo-audio`, `expo-image`, `expo-sharing` config plugins** — required by the updated package versions (`expo install --fix` refuses to auto-write through the dynamic `app.config.js`).
- **`react-native-track-player` added to `reactNativeDirectoryCheck.exclude`** — the library has no New Architecture support upstream; replacing it is out of scope for this PR, and the warning is otherwise unactionable.

**Remaining doctor failure (deliberate):** `Check that @react-navigation packages are not installed alongside expo-router` still fails: `@react-navigation/material-top-tabs` and `@react-navigation/native` remain direct dependencies. The app code already imports exclusively through expo-router re-exports (`expo-router/js-top-tabs`, `expo-router/react-navigation`), so removal looks safe on paper, but it needs runtime verification on device/emulator first and is left for a follow-up.

## 🏷️ Ticket / Issue

N/A

N/A — no UI changes.

## ✅ Checklist

- [x] Verified that changes behave as expected for all platforms (CI builds pass: iOS signed + unsigned, tvOS, Android phone + TV)

## 🔍 Testing Instructions

1. `bun i`
2. `bunx tsc --version` → reports 6.0.3
3. `bun run typecheck` → passes with no errors (pre-existing `utils/jellyseerr` third-party errors remain filtered by the script, unchanged)
4. `bunx expo-doctor` → 20/21 checks pass (only the deliberate @react-navigation check fails, see above)
5. `bunx expo install --check` → "Dependencies are up to date"
6. `bun run prebuild` (and `bun run prebuild:tv`) → prebuild succeeds with the three added config plugins
7. Optional: revert the `tsconfig.json` `types` line and re-run `bun run typecheck` to reproduce the 20 errors caused by the TS 6.0 `types` default change


Gauvino:
@coderabbitai full review

app.json
```diff
@@ -107,6 +107,9 @@
       ],
       "expo-localization",
       "expo-asset",
+      "expo-audio",
+      "expo-image",
+      "expo-sharing",
       [
```
Simon-Eklundh, line 113: Would you happen to know the reason for this addition?

retardgerman (approved):

app.json
Gauvino: Yeah those three (expo-audio, expo-image, expo-sharing) all ship an Expo config plugin, and they were already installed as deps but never listed in the plugins array, so expo prebuild wasn't actually applying their native config. Adding them is mostly the recommended SDK 56 setup, but honestly the effect is different for each one.

expo-audio is the one that actually matters. Its plugin is what turns on background audio (UIBackgroundModes audio on iOS, the media playback foreground service on Android) plus the audio permissions, so it's there so background playback behaves correctly.

expo-image basically just sets a Podfile property (the libdav1d toggle) to its default value, so it's harmless either way.

expo-sharing is close to a no op the way it's listed. The share extension and Android intent filter stuff in its plugin has to be turned on explicitly via ios.enabled / android.enabled, both default false, so a bare "expo-sharing" entry doesn't add anything native. We only use it to export the logs file through the share sheet. I left it in just to keep the installed config plugins listed, but I'm totally fine dropping it if you'd rather keep app.json minimal.

Simon-Eklundh: Alright, that explains it!

Simon-Eklundh (approved):

PR #1783, opened 2026-07-01
feat: adding exoplayer for HDR playback

lancechant:
# 📦 Pull Request

[AI Assisted](#) 
Used for the structure and layout, the rest was fixed and added and tested manually

### Summary

  Adds AndroidX Media3 (ExoPlayer) as an alternative video player on Android TV, accessible via a new TV settings selector. MPV remains the default on every platform; ExoPlayer is opt-in. The two players share
   an identical TypeScript contract (MpvPlayerViewRef) so the rest of the app swaps between them transparently.

###  Why

  MPV on Android TV currently tone-maps HDR content to SDR — no target-colorspace/target-trc is wired into the gpu-next pipeline. ExoPlayer uses MediaCodec directly, so HDR10/HDR10+ metadata reaches the
  surface layer and the display correctly engages HDR mode. ExoPlayer is also materially lighter on low-RAM Android TV boxes than the libmpv 1.0 path, which was causing(But should be reduced now) OOM during 4K HDR playback.

  
### Skip-intro UX fixes (apply to both players — control layer, not ExoPlayer-specific)

  - TVSkipSegmentCard + TVNextEpisodeCountdown: BOTTOM_WITH_CONTROLS was raw 300 and unscaled; reduced to scaleSize(220) so the card sits just above the controls bar instead of floating ~100px above it.
  Mirrored in both components so the mutually-exclusive cards don't pop.
  - Controls.tv.tsx: wrapped skipIntro / skipCredit onPress to call setFocusPlayButton(true) before the seek — otherwise the card unmounts after the seek and the focus engine is stranded when controls are
  visible.

## ✅ Checklist

- [ ] Code passes lint/formatting and type checks (`tsc`/`biome`)
- [ ] No secrets, hardcoded credentials, or private config files are included
- [ ] I've declared if AI was used to assist with this PR (by uncommenting the line at the bottom, or not)

## 🔍 Testing Instructions

Test plan

  - Android TV: open settings, verify "Video Player" selector appears under Audio section
  - Select ExoPlayer → confirm ExoPlayer note appears, MPV note disappears
  - Select MPV → confirm MPV note appears, ExoPlayer note disappears
  - Verify VO driver + Horizontal Alignment rows hide when ExoPlayer is selected
  - Verify subtitle scale / font size / vertical margin / vertical alignment / background rows stay visible for both players
  - Play HDR10 content with ExoPlayer selected → TV's HDR badge engages, technical overlay shows HDR10 / BT.2020
  - Play 4K HEVC HDR10+ content → no purple/green cast (DV Profile5 excluded by device profile)
  - Play a DTS or TrueHD track → audio plays (FFmpeg software decode → multichannel PCM to sink), technical overlay shows Decoder: FFmpeg… (software)
  - Play SDR HEVC content → technical overlay shows Decoder: c2.amlogic.hevc.decoder (hardware), no Color row (SDR has no color metadata to show)
  - Trigger skip-intro with controls visible → focus lands on play button, not stranded
  - iOS / Android mobile → no regressions; ExoPlayer selector hidden, MPV still default

## Summary

* **New Features**
  * Added an Android TV–only **Video Player** selector (ExoPlayer or MPV) with backend-specific limitation notes, plus MPV-only options shown only when MPV is selected.

* **Bug Fixes**
  * Improved TV focus behavior after intro/credits skipping.
  * Fixed time/remaining-time formatting for invalid values and removed initial Infinity artifacts.
  * Enhanced Android TV MPV resume recovery while preserving external subtitles.

* **UI Improvements**
  * Refined TV layout spacing for next/skip cards.

* **Improvements**
  * Expanded the technical info overlay with additional player-reported details.


components/video-player/controls/TechnicalInfoOverlay.tsx
```diff
@@ -339,6 +347,17 @@ export const TechnicalInfoOverlay: FC<TechnicalInfoOverlayProps> = memo(
...
             </Text>
           )}
+          {(info?.colorSpace || info?.colorRange || info?.colorTransfer) && (
+            <Text style={textStyle}>
+              Color:
+              {[info.colorSpace, info.colorRange, info.colorTransfer]
+                .filter(Boolean)
+                .join(" / ")}
+            </Text>
+          )}
```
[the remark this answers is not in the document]
Gauvino: Maybe also need to add translation key for it

modules/exoplayer-player/android/src/main/java/expo/modules/exoplayerplayer/ExoPlayerView.kt
```diff
@@ -0,0 +1,905 @@
...
+
+    // MARK: - Video Loading
+
+    fun loadVideo(config: VideoLoadConfig) {
+        if (currentUrl == config.url) return
+        currentUrl = config.url
+        pendingConfig = config
+        ensurePlayer(config)
+        loadInternal(config)
+    }
```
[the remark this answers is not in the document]
lancechant: This is the same behavior with MPV, it's to stop reloading if using the same url is already playing

utils/profiles/native.ts
```diff
@@ -126,6 +170,63 @@ const getVideoAudioCodecs = (
...
+              Condition: "NotEquals",
+              Property: "VideoRangeType",
+              // Blocks ONLY pure DV Profile 5 (IPTPQc2). Profiles 7/8 with
+              // HDR10/SDR base layers fall through to Media3's HEVC fallback
+              // (1.9.1+). See getExoPlayerDirectPlayProfile doc above.
+              Value: "DOVI",
+              IsRequired: true,
+            },
+          ],
+        },
```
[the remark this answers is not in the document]
lancechant: This is what we are doing above for mpv, and stops only the profile 5

modules/mpv-player/android/src/main/java/expo/modules/mpvplayer/MPVLayerRenderer.kt
```diff
@@ -142,6 +142,13 @@ class MPVLayerRenderer(private val context: Context) : MPVLib.EventObserver {
      */
     private var voDriver: String = "gpu-next"
 
+    /**
+     * True on Android TV form-factor devices. Drives both the hwdec selection
+     * in [start] and the TV-only resume-recovery gating in [MpvPlayerView].
+     * Computed once from the system UI mode.
+     */
+    val isTv: Boolean = isTvDevice()
+
```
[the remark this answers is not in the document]
lancechant: this is fixed now

modules/exoplayer-player/android/src/main/java/expo/modules/exoplayerplayer/ExoPlayerView.kt
```diff
@@ -0,0 +1,976 @@
...
+        }
+
+        override fun onIsPlayingChanged(isPlaying: Boolean) {
+            onPlaybackStateChange(mapOf(
+                "isPlaying" to isPlaying,
+                "isPaused" to !isPlaying
+            ))
+        }
+
+        override fun onPlayerErrorChanged(error: androidx.media3.common.PlaybackException?) {
```
Gauvino, line 203: This also fires when the error clears, with error = null (for example after a new prepare() from addSubtitleFile or a retry). In that case we send "Unknown playback error" to JS while playback is actually recovering. Probably worth an early return when error is null.

modules/exoplayer-player/android/src/main/java/expo/modules/exoplayerplayer/ExoPlayerView.kt
```diff
@@ -0,0 +1,976 @@
...
+
+    fun setSpeed(speed: Double) {
+        player?.playbackParameters = PlaybackParameters(speed.toFloat())
+    }
+
+    fun getSpeed(): Float {
+        return player?.playbackParameters?.speed ?: 1f
+    }
+
+    fun isPaused(): Boolean {
```
Gauvino, line 502: isPlaying is also false while buffering, so this reports paused during a rebuffer even when playWhenReady is true. MPV returns the pause property here. `player?.playWhenReady == false` would match the mpv behavior better.

utils/profiles/native.ts
```diff
@@ -126,6 +170,63 @@ const getVideoAudioCodecs = (
...
+        {
+          Type: MediaTypes.Audio,
+          Codec: "vorbis,opus,flac,alac,pcm,mp3,aac,ac3,eac3,dts,dtshd,truehd",
+        },
+      ],
+      DirectPlayProfiles: [exoDirect.video, exoDirect.audio],
+      TranscodingProfiles: [
```
Gauvino, line 209: The mpv profile also has an audio TranscodingProfile (mp3/http) as a fallback, the exo branch only has the video hls one. If an audio item ends up on this profile and can't direct play, the server has no transcode target. Might be worth adding the same audio entry for parity.

modules/exoplayer-player/android/src/main/java/expo/modules/exoplayerplayer/ExoPlayerView.kt
```diff
@@ -0,0 +1,976 @@
...
+        return subtitleTrackList.map { entry ->
+            mapOf(
+                "id" to entry.id,
+                "title" to (entry.format.label ?: ""),
+                "lang" to (entry.format.language ?: "")
+            )
+        }
+    }
+
+    fun setSubtitleTrack(trackId: Int) {
```
Gauvino, line 607: Small defensive thing: the entries come from the last onTracksChanged snapshot, and after addSubtitleFile rebuilds the MediaItem the group indices can be stale for a moment. `groups.getOrNull(entry.trackGroupIndex) ?: return` would avoid a possible IndexOutOfBounds. Same for setAudioTrack below.

modules/exoplayer-player/android/src/main/java/expo/modules/exoplayerplayer/ExoPlayerView.kt
```diff
@@ -0,0 +1,976 @@
...
+        val backBufferMs = if (!cacheEnabled) {
+            0
+        } else {
+            val mb = config.demuxerMaxBackBytes ?: 50
+            // Heuristic: 1 MB ≈ 1s of typical 1080p bitrate.
+            (mb * 1000).coerceAtLeast(1000)
+        }
+
+        val builder = DefaultLoadControl.Builder()
+            .setTargetBufferBytes(if (!cacheEnabled) 0 else ((config.demuxerMaxBytes ?: 150) * 1024 * 1024))
```
Gauvino: When cacheEnabled is false this ends up calling setTargetBufferBytes(0). Not sure how DefaultLoadControl behaves with a 0 byte target, did you test with cache mode set to "no"? If not, C.LENGTH_UNSET might be safer there.

Gauvino (changes requested):
Went through the whole diff, really nice work overall. The player swap architecture is clean and the resume recovery on mpv is a welcome bonus. A few small things on the exo side, mostly error handling and buffering edge cases, details inline.

utils/profiles/native.ts
lancechant: Thanks, I think I was testing and forgot to add it.. as was on my local branch

lancechant:
> Went through the whole diff, really nice work overall. The player swap architecture is clean and the resume recovery on mpv is a welcome bonus. A few small things on the exo side, mostly error handling and buffering edge cases, details inline.

Thanks for looking, have gone through and fixed the mentioned items

Gauvino (approved):

PR #1723, opened 2026-06-12
fix: pre-release bug-fix batch (filters, Quick Connect, auth, downloads, i18n)

Gauvino:
# 📦 Pull Request

[AI Assisted](#)

> 📋 **Per-platform test checklist (clickable):** see the [pinned checklist comment](https://github.com/streamyfin/streamyfin/pull/1723#issuecomment-4722907907) — Android validated, iOS pending on the signed IPA.

## 📝 Description

Pre-release bug-fix batch — one commit per fix. Branch synced with the latest `develop` (TypeScript 6 / SDK 56 alignment #1719, dead-code cleanup #1625, …).

| Fix | Summary |
|---|---|
| **Filters never opened** (new arch) | `BottomSheetModal.present()` from a `useEffect` silently no-ops on Fabric + Reanimated 4. Filter sheets now present imperatively from the press handler, open instantly with a loader; same fix for the bitrate/source/track sheets. |
| **Filter search ate keystrokes** | On large option lists (2000+ tags) each keystroke re-filtered + re-rendered the list, desyncing the controlled input (lost/reappearing letters). `useDeferredValue` + memoized rows. |
| **Quick Connect code stuck on screen** | The code lived in a native Alert (no programmatic dismiss). Now an in-app sheet that auto-closes once the session is authorized, with tap-to-copy (expo-clipboard). Polling also stops when leaving the login screen or going back to server selection. |
| **Save account asked before the login** | The protection picker showed before the credentials were even checked (and Quick Connect couldn't save at all). Login flows now flag the intent; a global modal asks AFTER the session is authorized — identical for password and Quick Connect. Cancelling saves nothing. |
| **Session not cleared on 401** | A response interceptor clears the session once when the server revokes the token — no more 401 spam / half-authenticated state. Saved accounts survive. |
| **Offline splash hang (75-120 s)** | Startup no longer awaits `getCurrentUser` (axios has no timeout); cached user renders immediately, validation runs in the background. |
| **Admin-locked settings changeable** | `updateSettings` strips plugin-locked keys at write time; the Streamystats screen shows the live admin value and disables locked controls. |
| **Delete all downloads: no confirmation** | Destructive action now asks first. |
| **Storage bar showed 0.00%** | Size computed from live files on disk, auto-refreshes every 10 s, refreshes right after delete-all, and in-progress downloads count as app usage. |
| **Recycled series poster** | Poster cache read keyed on `SeriesId` (was frozen on first render). |
[...]

Issue #138, opened 2024-09-26
Ability to run app in MacOS without turning off security level


rodrigogonegit:
**Describe the solution you'd like**
When trying to open the App in MacOS, the following is shown. I imagine it's the download feature that makes it so? Would it be possible not to?
<img width="297" alt="image" src="https://github.com/user-attachments/assets/7df308f8-2723-419e-acf6-dfd3b1217fa2">

**Additional context**
Relevant context: [in this Reddit post](https://www.reddit.com/r/MacOS/comments/18uvls7/comment/kfnhwwc/?context=3&share_id=50CWG4uQSmzdof1RdDa2o&utm_content=1&utm_medium=ios_app&utm_name=ioscss&utm_source=share&utm_term=1)

fredrikburmester:
That's strange, I just ran the app on my computer a couple of days ago.. Will look into this.

PR #1806, opened 2026-07-14
fix(nav): drop duplicate pushes from rapid taps

Gauvino:
# 📦 Pull Request

[AI Assisted](#)

## 📝 Description

| Fix | What was wrong | What changes |
|---|---|---|
| **Duplicate screen from rapid presses** | Two fast taps on the same poster open the item screen twice, the second identical to the first, so it takes two backs to get out. A push is not committed the moment it is called, it goes through expo-router's routing queue, so when the second press lands nothing has changed yet that would stop it. | `useAppRouter.push` drops a push from a screen that is already pushing. The screen counts as busy from the first push until it regains focus, and the flag is a synchronous ref rather than the focus state, so a duplicate inside the same JS batch is caught too. |
| **Same thing on Android TV, with a longer window** | On TV the poster keeps receiving remote presses for about a second after the screen it opened is already on top, so a second select stacks another copy. The row underneath is not visible any more, so that press is not something the user can see themselves making. | Same guard. It covers both shapes: the same-batch duplicate through the ref, and the late press through the focus state of the source screen. |
| **Keyboard stayed open on back** | Leaving a screen with the header back button left the keyboard over the previous screen, for instance leaving the Jellyseerr login while typing. | The header back button dismisses the keyboard before navigating. |

**Note on the original Android TV claim.** This PR first blamed a remote select firing `onPress` twice ([react-native-tvos#110](https://github.com/react-native-tvos/react-native-tvos/issues/110), [#138](https://github.com/react-native-tvos/react-native-tvos/issues/138)). Instrumenting the build shows that does not happen any more on react-native-tvos 0.86: one select gives exactly one `onPress`, and `Pressability` already filters the spurious Android TV `onClick`. The duplicate on TV comes from the same uncommitted-push window as on phone.

**Scope.** A press that lands on a poster hidden under the screen it just opened is dropped. Opening two different posters normally, one after the other, is untouched: going back and opening another one works in every run.

> Part of splitting #1723 into focused PRs.

## 🏷️ Ticket / Issue

Related: #1723 (split)

**iPhone, two fast taps on the same poster.**

| Before | After |
|---|---|
| two screens pushed | one screen pushed |

In the before clip the item screen slides in twice, the second over an identical first, and it takes two backs to return to the home screen. In the after clip the same gesture opens it once.

Measured from the recordings rather than read off them, since two identical screens are hard to tell apart by eye: the opening animation of the current build is one continuous block of 1.06s, which is two chained transitions, against roughly 0.5s per transition on this branch.

**Android TV, three posters selected in a row, then unwinding with back.**

| Before | After |
|---|---|
| three screens stacked | one screen |

Same key sequence in both, select, right, select, right, select. Before, the app ends up three screens deep and needs three backs to leave. After, only the first poster opens and one back returns home. The second and third presses land on the row hidden under the item screen, about a second after it opened.

Numbers from the instrumented build on the emulator, tight double press on the same poster, hard reset between runs:

| | Without the guard | With the guard |
|---|---|---|
| Screens stacked | 2, in 6 runs out of 6 | 1, in 6 runs out of 6 |
| One back after the presses | still on the item screen | back on the home screen |

Same result on the library grid, the series page, the hero carousel and the filter sheet: one screen each. Five open and back cycles in a row give five pushes and zero dropped, so the guard does not stick.

## ✅ Checklist

## 🔍 Testing Instructions

1. Tap a poster twice, as fast as you can. One screen opens, one back returns to the list
2. Tap a poster once, come back, tap it again. Both taps open the screen, the guard does not stick
3. Cold start the app and tap a poster as early as the home screen appears, then tap it again a moment later. The screen opens, nothing stays blocked
4. Focus a text input, for example the search field or the Jellyseerr login, then leave with the header back button. The keyboard closes with the screen
5. Normal navigation is unchanged: tabs, back, deep links
6. On Android TV, press select twice quickly on a poster. One screen opens, one back returns home
7. On Android TV, open a poster, go back, open another one. Both open
8. On Android TV from a series page, Shuffle then pick a pool in the sheet. The player still starts, the guard releases before that push

Tested on iOS on device and on Android TV on the emulator. The duplicate reproduces without this branch and does not with it, repeated navigation keeps working afterwards, and an early press after a cold start opens the screen normally.


hooks/useAppRouter.ts
```diff
@@ -19,24 +25,59 @@ export function useAppRouter() {
   const router = useRouter();
   const isOffline = useOfflineMode();
 
+  // Optional: undefined when used outside a navigator (root layout, providers).
+  // When present it reflects the focus state of the screen this hook lives in.
+  const navigation = useContext(NavigationContext);
+
+  // Synchronous re-entry guard for TV: a single remote "select" on Android TV
+  // can fire onPress more than once within the same JS batch
+  // (react-native-tvos#110/#138), BEFORE react-navigation commits the pushed
```
lancechant, line 34: These issues were/are fixed? or needed to be stopped with a platform check?

Gauvino: For now I tried hard to fix it on Android TV ut it didn't conclude, im on some fix will try it when back home or maybe tomorrow cause I don't have much time today

Gauvino: @lancechant look at the description it work's great !

lancechant (approved):

PR #1807, opened 2026-07-14
fix(settings): enforce admin locks and plugin defaults

Gauvino:
# 📦 Pull Request

[AI Assisted](#)

## 📝 Description
| Fix | Summary |
|---|---|
| **Admin-locked settings changeable** | `updateSettings` strips plugin-locked keys at write time; the Streamystats screen shows the live admin value and disables locked controls (enablement derived from the effective server URL). |
| **Unlocked plugin setting fell back to the app default** | A setting that was always locked then unlocked showed the app's hardcoded default instead of the plugin's configured default. The user's value now wins only when it diverges from the app default; otherwise the plugin (admin) default applies (shared `useSettings`, so mobile + TV). |
| **Jellyseerr lock too broad** | Only the server URL is admin-locked; the password stays editable. |
| **Marlin Search lock notice** | Shown via the standard ListItem subtitle instead of a custom layout. |
| **Kefin tweaks** | Uses the standard ListItem/disabledByAdmin pattern. |
| **Plugin list order** | Jellyseerr / Streamystats / Marlin restored. |
| **"Refresh from server"** | Moved to the plugins index page. |

⚠️ Release-note worthy: a user who had explicitly picked the app-default value gets the admin default when the setting is unlocked (the two states are indistinguishable).

> Part of splitting #1723 into focused PRs.

## 🏷️ Ticket / Issue
Related: #1723 (split)

**Locked settings (live admin value, disabled controls):**

locked settings

## ✅ Checklist

## 🔍 Testing Instructions
1. Lock a setting in the Streamyfin plugin (Jellyfin admin) → the app shows the admin value, control disabled, never persists local changes
2. Unlock it → editable again and persists; a setting that was ALWAYS locked shows the plugin default (not the app default) after unlock + "refresh server settings"; a user value that diverged from the app default survives the unlock (mobile AND TV — shared hook)
3. Jellyseerr plugin page: URL locked by admin, password still editable
4. Settings → Plugins: order is Jellyseerr / Streamystats / Marlin Search; "refresh from server" lives on the plugins index

Already device-tested on Android + iOS as part of #1723.

## Summary by CodeRabbit

* **New Features**
  * Added a “Refresh from server” action to the Plugins settings page, with localized error feedback when refresh fails.
  * Enhanced admin-locked plugin settings: they show administrator-provided values and are no longer editable.

* **Improvements**
  * Updated Jellyseerr, Kefin Tweaks, Marlin Search, and Streamystats settings UIs to use consistent, simplified locked/disabled states.
  * Adjusted which screens support refresh-from-server and refined the save behavior so locked values aren’t persisted.
  * Updated settings list ordering and internationalized “Disabled by admin” messaging.

### DIFF
--- a/app/(auth)/(tabs)/(home)/settings/plugins/kefinTweaks/page.tsx
+++ b/app/(auth)/(tabs)/(home)/settings/plugins/kefinTweaks/page.tsx
@@ -1,11 +1,8 @@
-import { ScrollView } from "react-native";
+import { ScrollView, View } from "react-native";
 import { useSafeAreaInsets } from "react-native-safe-area-context";
-import DisabledSetting from "@/components/settings/DisabledSetting";
 import { KefinTweaksSettings } from "@/components/settings/KefinTweaks";
-import { useSettings } from "@/utils/atoms/settings";
 
 export default function KefinTweaksPage() {
-  const { pluginSettings } = useSettings();
   const insets = useSafeAreaInsets();
 
   return (
@@ -16,12 +13,9 @@ export default function KefinTweaksPage() {
         paddingRight: insets.right,
       }}
     >
-      <DisabledSetting
-        disabled={pluginSettings?.useKefinTweaks?.locked === true}
-        className='p-4'
-      >
+      <View className='px-4'>
         <KefinTweaksSettings />
-      </DisabledSetting>
+      </View>
     </ScrollView>
   );
 }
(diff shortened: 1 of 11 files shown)


app/(auth)/(tabs)/(home)/settings/plugins/kefinTweaks/page.tsx
```diff
@@ -1,11 +1,8 @@
-import { ScrollView } from "react-native";
+import { ScrollView, View } from "react-native";
 import { useSafeAreaInsets } from "react-native-safe-area-context";
-import DisabledSetting from "@/components/settings/DisabledSetting";
```
lancechant, line 3: So is this no longer used? I see it removed a lot, but not a full delete?

PR #1823, opened 2026-07-19
fix: playback stopped reporting every 30 s

Simon-Eklundh:
# 📦 Pull Request

[AI Assisted](#)
used for finding the cause and suggesting a quick fix.
## 📝 Description
fixes the playbackreporting stopped behaviour so it only happens once playback ends (and player exits) instead of once every 30 seconds.

## 🏷️ Ticket / Issue
N/A, I found it myself

N/A it's just a logs thing

## ✅ Checklist

## 🔍 Testing Instructions

before installing this version:
1. open jellyfin logs (ex. docker logs jellyfin -f)
2. start playback on something in streamyfin (develop)
3. see that every 30 seconds, playback reports an end

after installing this version:
repeat same steps except no playback end reports while playback is ongoing

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate “playback stopped” reports when leaving the player by deduplicating per stream session.
  * Improved teardown behavior so reporting happens only on real unmount, while navigation changes no longer trigger the same report.
  * Added safer error handling for playback-stop reporting (errors are logged and cleanup continues).
* **Chores**
  * Removed debug console logging related to filter options computation.

### DIFF
--- a/app/(auth)/player/direct-player.tsx
+++ b/app/(auth)/player/direct-player.tsx
@@ -472,24 +472,42 @@ export default function DirectPlayerPage() {
     }
   };
 
+  // PlaySessionId of the last "stopped" report, to dedupe the double teardown.
+  const reportedStopSessionIdRef = useRef<string | null>(null);
+
   const reportPlaybackStopped = useCallback(async () => {
     if (!item?.Id || !stream?.sessionId || offline || !api) return;
+    // Leaving the player reports "stopped" twice: once from the beforeRemove
+    // listener (via stop()) and once from the unmount cleanup backstop. Dedupe
+    // per PlaySessionId — not a plain boolean, since the component survives
+    // in-place item switches and the next session must report again.
+    if (reportedStopSessionIdRef.current === stream.sessionId) return;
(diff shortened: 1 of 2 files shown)


lancechant (dismissed):

lancechant:
I think the code rabbit comments just need to be checked and confirmed as they seem valid

Simon-Eklundh:
> I think the code rabbit comments just need to be checked and confirmed as they seem valid

I agree

So I've changed that part

fredrikburmester:
does this send playback stopped on next episode? looks like it only does that on screen unmount. Can you double check?

Simon-Eklundh:
> does this send playback stopped on next episode? looks like it only does that on screen unmount. Can you double check?

dang it

I forgot completely about next episode being a thing..... despite that being my most common playback "way"

I'll fix it tomorrow

Simon-Eklundh:
today is tomorrow I guess 

I fixed it (added Iitem?.id etc to dependency array (I think that's what it's called??) so it fires on media item swap as well as unmount and not just on unmount

lancechant (approved):

PR #1838, opened 2026-07-20
chore(deps): migrate to Expo SDK 57 (React Native 0.86)

Gauvino:
# 📦 Pull Request

[AI Assisted](#)

## 📝 Description

PR 7 of #1831: **Expo SDK 56 → 57 migration (React Native 0.85 → 0.86)**.

SDK 57 is a small release: RN 0.86 has no documented user-facing breaking changes, React stays 19.2.3, and every toolchain requirement (Xcode 26.4, AGP 8.12, compile/target SDK 36, Gradle 9.3.1) is identical to SDK 56 — verified against the official changelog, RN release blog, and the RN gradle version catalogs.

**Core bumps**

| Package | From | To |
|---|---|---|
| `expo` (+ all `expo-*` / `@expo/*`) | ~56.x | **~57.x** (expo 57.0.7) |
| `react-native` (react-native-tvos fork) | 0.85.3-0 | **0.86.0-2** (aligned with RN 0.86.0 stable, published 2026-06-26) |
| `expo-router` | 56.2.x | 57.0.7 (no breaking changes in the 57 line) |
| `react-native-screens` | 4.25.2 | 4.26.2 |
| `react-native-gesture-handler` | ~2.31.1 | **~2.32.0** (v2 line, carries the RN 0.86 support — v3 is deliberately deferred, see #1831 backlog) |

**Compatibility pairs (must move together — the reason the old standalone Renovate PRs #1381/#1419 failed CI for months):**

| Pair | To | Note |
|---|---|---|
| `react-native-reanimated` + `react-native-worklets` | **4.5.2 + 0.11.1** | One step past Expo's default pins (4.5.0/0.10.0) but inside the official compatibility matrix; 4.4.2 specifically fixed "freezing animations after RN 0.86", and worklets 0.11 ships the bytecode option relevant to the Hermes memory workaround |
| `react-native-mmkv` + `react-native-nitro-modules` | **4.3.2 + 0.36.1** | mmkv ≥4.2 hard-requires nitro ≥0.35; mmkv is our only Nitro consumer |

[...]

PR #1763, opened 2026-06-23
fix: subtitle ordering

lancechant:
# 📦 Pull Request

[AI Assisted](#)
Used just to bounce ideas off, logic was hand written and tested

## 📝 Description

Fixed an issue where external and subrip subtitles were not ordered correctly

## 🏷️ Ticket / Issue

#954 

## ✅ Checklist

- [X ] I’ve read the [contribution guidelines](CONTRIBUTING.md)
- [X ] Verified that changes behave as expected for all platforms
- [X ] Code passes lint/formatting and type checks (`tsc`/`biome`)
- [ X] No secrets, hardcoded credentials, or private config files are included
- [X ] I've declared if AI was used to assist with this PR (by uncommenting the line at the bottom, or not)

## 🔍 Testing Instructions

1. select external or subrip subtitles, it should now display correctly

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed subtitle track mapping to correctly reflect media player track ordering, ensuring subtitles display in the proper sequence when multiple subtitle streams are present.
  * Improved handling of various subtitle delivery methods while properly excluding unavailable subtitle types during transcoding operations.

PR #1847, opened 2026-07-21
fix(ios): keep subtitle/track menus in the provided order

Gauvino:
# 📦 Pull Request

[AI Assisted](#)

## 📝 Description

On iOS the subtitle selector (item page sheet and in-player track menus) shows its entries in the wrong order, while Android shows the same episode correctly. The data is sorted identically on both platforms (`compareTracksForMenu`, jellyfin-web order; #1763/#1781 are not involved): the difference is in rendering. Android uses a JS bottom sheet that respects array order. iOS renders through `@expo/ui`'s native SwiftUI `Menu`, and UIMenu reorders items by proximity to the anchor, so a menu opening upward (both call sites sit at the bottom of the screen) displays the list reversed.

Fix in three parts:

1. **`patches/@expo%2Fui@57.0.7.patch`**: apply SwiftUI's `.menuOrder(.fixed)` (iOS 16+, availability-guarded) to `MenuView`, so every `PlatformDropdown` menu renders items in the order the app provides. `menuOrder` is not exposed through `@expo/ui`'s modifier list, hence the package patch rather than app code. I opened expo/expo#48048 and expo/expo#48049 to get the modifier exposed upstream, which would let us drop the patch later.
2. **Pin `@expo/ui` to `57.0.7`** (was `~57.0.7`): `patchedDependencies` keys match exact versions, so a lockfile refresh drifting to a newer release would orphan the patch. Pinned, any bump becomes a deliberate `package.json` edit that the guard below validates in the same PR.
3. **`scripts/check-patched-deps.ts` + a step in the lockfile workflow**: bun only applies a patch whose `patchedDependencies` key matches the exact resolved version and *silently skips it otherwise* (version-less keys are ignored entirely, verified empirically). Without a guard, the next `@expo/ui` bump would silently drop this fix. The check parses `bun.lock` with `Bun.JSONC` and fails CI with regeneration instructions instead.

## 🏷️ Ticket / Issue

N/A, found while device-testing #1838 (the ordering bug predates the SDK bump). Upstream issue: expo/expo#48048.

Before (iOS): externals listed first / list reversed in the subtitle sheet. 

image

After: same order as Android and jellyfin-web (embedded first, externals last). Device screenshots to follow after the next iOS build.

image

## ✅ Checklist

- [ ] Verified that changes behave as expected for all platforms (iOS device test pending, needs a fresh iOS build with the patched pod)
- [x] I've declared if AI was used to assist with this PR (badge above)

## 🔍 Testing Instructions

1. Build the iOS app (the ExpoUI pod compiles from the patched `node_modules`).
2. Open an episode with several external subtitles, then open the subtitle selector on the item page: order must match Android / jellyfin-web (embedded first, then externals, forced/default up).
3. In the player, open the subtitle and audio menus: same order top-to-bottom as Android.
4. Guard check: run `bun scripts/check-patched-deps.ts`, it passes. Edit the `patchedDependencies` key to a wrong version and it fails with instructions (CI runs this in the lockfile workflow).
5. Android/TV regression: menus unchanged (they don't go through `@expo/ui`'s Menu).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved menu behavior by preserving menu item order across supported Apple platforms.
  * Updated menu interactions so primary actions and long-press menu access work consistently.

* **Chores**
  * Added validation to detect missing or incorrectly applied dependency patches during CI checks.


fredrikburmester:
Can we use @expo/ui 57 on Expo 55? Should we update The Expo version first?

patches/@expo%2Fui@57.0.7.patch
```diff
@@ -0,0 +1,65 @@
...
+-        } else if let labelContent {
+-          Menu { ChildrenWithoutLabel() } label: { labelContent } primaryAction: { props.onPrimaryAction() }
+-        } else if let label = props.label {
+-          Menu(LocalizedStringKey(label)) { Children() } primaryAction: { props.onPrimaryAction() }
+-        }
++      if #available(iOS 16.0, tvOS 17.0, *) {
++        // Keep items in the order they were provided: without this, UIMenu
++        // reorders entries by proximity to the anchor, so a menu that opens
++        // upward shows the list reversed (selection menus end up backwards).
++        menuContent().menuOrder(.fixed)
```
Simon-Eklundh, line 52: is it possible we ever don't want .fixed?

possibly somewhere else than in the subtitle list?

fredrikburmester (approved):

patches/@expo%2Fui@57.0.7.patch
Gauvino: Every menu we render through @expo/ui is a selection list (subtitle and audio tracks, scaling modes, mpv settings, media sources), and for those the declaration order is always the order we build in JS, so I can't think of a case where we'd want the default back. The default isn't a smarter ordering by the way, UIMenu just reverses the entries when the menu opens upward so the first one sits near the finger. That makes sense for action menus, but for selection lists it reads as broken, which is exactly the bug this fixes.

If we ever add an action style menu where the adaptive order is wanted, I'd extend the patch to expose menuOrder as a prop with fixed as the default instead of hardcoding it. I'm also pushing this upstream (expo/expo#48049), so the patch should be temporary anyway.

PR #1854, opened 2026-07-22
fix(music): tie the queue and playback to the signed-in account

Gauvino:
# 📦 Pull Request

[AI Assisted](#)

## 📝 Description

The music queue lived in MMKV (`music_player_queue`) with no notion of who it belonged to, so it followed you across sessions: after connecting to another server, the mini player restored tracks from the previous one, whose item ids and stream URLs mean nothing on the new server, and playback could simply keep going.

This ties the queue and the player to the signed-in account:

- When the session moves to another server, another user, or logs out, the provider resets TrackPlayer, the in-memory state and the persisted queue. Repeat and shuffle are kept, they are user preferences rather than server state.
- The persisted queue is now stamped with its owner (`serverId:userId`) and is dropped on restore when that owner does not match, so another user on the same server no longer inherits it. A restored queue is also rejected when any of its items belongs to another server, not just the current one.
- The queue is no longer written back right after the session teardown. That save ran once more with the pre-reset queue, which put it back in storage with no owner on logout, or under the new owner on an account switch.
- Work already in flight carries a session generation: `loadAndPlayQueue`, the background loader and the look-ahead check it after every await, so a load started for the previous account cannot re-add its tracks behind the reset or report a playback start to the old server.
- Storage cleanup is best effort per key, and runs in a `finally` so a failing `TrackPlayer.reset()` cannot leave a queue that restores on the next launch.

TV is unaffected, its provider is a no-op stub.

## 🏷️ Ticket / Issue

N/A (reported on Discord)

Play a track on server A, kill the app, then switch to server B.

| Before | After |
|---|---|
| before | after |

Before, the mini player comes back on server B with "Fire" by 2NE1, a track that only exists on server A, and it covers the bottom row of the home screen. After, nothing is restored and the row stays visible.

| Before | After |
|---|---|
| before | after |

## ✅ Checklist

## 🔍 Testing Instructions

Server A needs a music library, server B does not have to.

1. On server A, play a track, then minimize the player so the mini player bar is visible
2. Switch to server B: the mini player must be gone and no audio from server A must keep playing
3. Kill and relaunch the app on server B: the mini player must not come back with server A's queue
4. Back on server A: music works normally and starts from an empty queue
5. Same server, other user: log out, log in as another account, the previous account's queue must not be restored
6. Same account: play something, kill the app, relaunch, the queue is restored as before
7. Log out entirely while music is playing: playback stops and the notification is cleared


lancechant (approved):

fredrikburmester:
Good improvement!

PR #1873, opened 2026-07-26
fix(player): report playback stopped when exiting a downloaded item

Alexk2309:
# 📦 Pull Request

[AI Assisted](#)

## 📝 Description

Exiting a downloaded item never closed its server-side session, so it stayed listed as active on the Jellyfin dashboard.

`reportPlaybackStopped` returned early whenever `offline` was set, but that flag only means the source is a local file — it says nothing about whether the device has a network connection. Progress *is* still sent to the server in that state, because `usePlaybackManager.reportPlaybackProgress` posts whenever it has a connection. So Jellyfin opened a session, received progress, and then never got a stop.

Now gated on the actual connection instead. Two smaller things had to change with it:

- Downloaded streams are constructed with `sessionId: ""`, which also tripped the `!stream?.sessionId` guard. The dedupe key falls back to the item id when there's no PlaySessionId.
- `PlaySessionId` is omitted from the payload rather than sent as an empty string.

Genuinely disconnected playback is unchanged: `isConnected` is false, so it returns early exactly as before.

## 🏷️ Ticket / Issue

N/A

N/A — no UI changes. Visible in the Jellyfin dashboard's active session list.

## ✅ Checklist

## 🔍 Testing Instructions

Open **Dashboard → Active Devices** in the Jellyfin web UI alongside the app.

**Downloaded item, with network**
1. Download an item, then play it from the downloads screen with wifi on
2. Confirm the session appears in Active Devices
3. Exit the player
4. Confirm the session disappears, and the resume position on the server matches where you stopped

**Downloaded item, no network**
1. Enable airplane mode and play a downloaded item
2. Exit the player
3. Confirm no crash, no error toast, and the local resume position is still saved

**Streamed item (regression)**
1. Play a normal streamed item and exit
2. Confirm the session still closes as before

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved playback “stopped” updates to the playstate service so they only occur when network connectivity is available.
  - Avoided sending unnecessary stop notifications while offline.
  - Reduced duplicate “stopped” notifications by using a more reliable deduping key for both session-based and session-less playback.
  - Updated stop reporting so the play session identifier is sent when available, and cleared appropriately on matching failures.

### DIFF
--- a/app/(auth)/player/direct-player.tsx
+++ b/app/(auth)/player/direct-player.tsx
@@ -37,6 +37,7 @@ import {
 import { VideoPlayerView } from "@/components/video-player/VideoPlayerView";
 import useRouter from "@/hooks/useAppRouter";
 import { useHaptic } from "@/hooks/useHaptic";
+import { useNetworkStatus } from "@/hooks/useNetworkStatus";
 import { useOrientation } from "@/hooks/useOrientation";
 import { usePlaybackManager } from "@/hooks/usePlaybackManager";
 import usePlaybackSpeed from "@/hooks/usePlaybackSpeed";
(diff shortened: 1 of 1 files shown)


Gauvino (approved):

Gauvino:
@Alexk2309 please check on this : https://github.com/streamyfin/streamyfin/pull/1879

Issue #761, opened 2025-06-09
Playback stops Android


20Bosko07:
Hello
I have the problem that the playback on Android only runs for 10sec and then stops. You need to restart the stream then. but can only wahvh 10 sec.

Is this a possible bug or a problem on my side?

Simon-Eklundh:
Hi there! That might be an issue on your side (doesn't happen to me, also on android)

But it is interesting 

What I'd recommend is checking network speed, connection to your server, speed, and server CPU/GPU/ram usage as a start

Also: what kind of media?

20Bosko07:
Hey @Simon-Eklundh 

I now tested a little bit around. On my iPad the playback works very well. Skipping forward, back, to the middel, etc. of a movie with 0 problems. On my phone (Google Pixel 7 Pro) I can start the playback without any problems, but under certain conditions when skipping forward and back the playback stops, and you only can restart it, when exiting and reopen the movie. Sometimes the control buttons disappear then. I dont think it is a problem on server side, so streaming in Jellyfin App and Browser works fine on all devices. And yes, it is 4K content. It's like the playback is not stable on Android. One thing maybe: My phone has left at the moment 5GB storage left, maybe is this the problem. I already tried to free 20gb of storage, same problem...

But can I provide to you something what can help to fix the problem?

Simon-Eklundh:
I experienced something similar; after 16 minutes, playback stopped

but I couldn't replicate it so idk....

TilDoesCode:
I dug into this and I believe I found the root cause: mpv (the default renderer on every platform here) has no native HTTP client — every remote stream is read through FFmpeg's `http`/`https` protocol, whose auto-reconnect is **off by default**. Neither `MPVLayerRenderer.swift` nor `MPVLayerRenderer.kt` ever set `network-timeout`/`stream-lavf-o`, so when the TCP connection dies mid-stream (proxy idle-timeout, cellular handover, brief server hiccup — all things that get more likely the longer/more mobile a session is, which fits "runs only ~10s on cellular" and "works fine on iPad/stable network but not on phone" from this thread), the demuxer's read just fails and playback silently stops advancing with no error — matching "you need to restart the stream then."

Opened a fix with a reproduction and regression test: #1878

PR #1878, opened 2026-07-27
fix(mpv): reconnect automatically when a stream's TCP connection drops

TilDoesCode:
## Summary

Fixes the "stream stops after a while, only seeking or restarting helps" playback stall (reported in #761 and consistent with the intermittent pattern in that thread).

- mpv has no native HTTP client — every remote source (DirectPlay, DirectStream, and each HLS transcode segment) is read through FFmpeg's `http`/`https` protocol via `avio_open2()`.
- FFmpeg's reconnect behavior for that protocol is **off by default**. Neither `MPVLayerRenderer.swift` (iOS/tvOS) nor `MPVLayerRenderer.kt` (Android) ever set `network-timeout` or `stream-lavf-o`.
- So when the underlying TCP connection dies mid-stream (reverse-proxy idle-timeout, cellular handover, Wi-Fi/cellular switch, a brief server hiccup), the demuxer's read just fails and mpv silently stops advancing — no error, no retry. A seek reopens the stream at a new offset (a new connection), which is why seeking "fixes" it; a full stop/restart obviously gets a fresh connection too.
- Fix: set `network-timeout=10` and `stream-lavf-o=reconnect=1,reconnect_streamed=1,reconnect_on_network_error=1,reconnect_delay_max=5` once at mpv start-up on both renderers, so every load path is covered.
- `reconnect_at_eof` is deliberately **not** set — enabling it made a real `avio_read()` loop retry for several seconds past a correct, complete end-of-stream before giving up, which would delay the natural end of every video. Full writeup and reproduction: `modules/mpv-player/NETWORK_RESILIENCE.md`.

## Verification

Built a minimal C harness against the same `libavformat`/`libavutil` this app links, calling `avio_open2()` with the same options mpv forwards — the exact layer mpv's `stream_lavf.c` uses. A local HTTP server serves a real MPEG-TS file and RSTs the connection ⅓ of the way through (not a clean close), then serves correctly on retry:

- **Without the fix**: read stops at 113,740 / 341,220 bytes, I/O error — reproduces the freeze.
- **With the fix**: auto-reconnects via a `Range` request at the drop point, full 341,220 bytes, done in 0.33s.
- **Regression check**: identical byte-for-byte result on a healthy (non-flaky) connection — no behavior change for the normal case.

Added `modules/mpv-player/networkResilience.test.ts` (runs under the project's real `bun:test`) asserting both native renderers carry the exact verified option set, applied before `mpv_initialize`/`mpv.initialize()`, with `reconnect_at_eof` absent. Confirmed it fails against the pre-fix source (6/6 fail) and passes against the fix (6/6 pass) — not a tautology.

## Test plan

- [x] `bun test` — 53/53 pass (repo-wide, including the new suite)
- [x] `bun run typecheck` — 0 errors
- [x] `bun run lint` (biome) — 644 files, clean
- [x] `bun run i18n:check` — clean
- [x] New test verified to fail on unpatched source, pass on patched source
- [ ] Real-device field test against a flaky/proxied Jellyfin connection (I don't have access to a reproducing server/network from here — would appreciate a test from someone who can reproduce #761 reliably)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved remote/HTTP(S) playback resilience on iOS and Android by enabling libmpv network reconnect behavior.
  * Reduced mid-stream freezes by configuring mpv timeout and ffmpeg HTTP reconnect settings (with reconnect delay limits).

* **Tests**
  * Added a Bun regression test to verify renderers set the required mpv network resilience options and that reconnect-at-EOF is intentionally disabled.

* **Documentation**
  * Added a guide describing the freeze cause and how the reconnect configuration addresses it.

* **Chores**
  * Updated TypeScript exclusions to properly account for the Bun test file.


lancechant:
Hey @TilDoesCode this issue was old, and on the old player.. is this still an issue currently? were you able to reproduce before? please also follow the PR template, and show proof you have been able to replicate before and the fix after works

PR #1888, opened 2026-07-28
fix(tv ui): Display tv show name on episodes on home screen

### LINKED ISSUE #1886: [Bug]: TV show name is missing on TV app in main screen

### What happened?

When scrolling the home on tv, all the episodes are missing the tv show name.
The only way to know which show is by the thumbnail.

I want the tv show name to be shown for episodes on the TV app.

### Reproduction steps

1. Open streamyfin app on tvOS.
2. notice that the episodes don't tell which tv show they belong to.

Apple TV 4k tvOS 26.5

### Streamyfin Version

0.54.1

- Jellyfin Server Version: 10.11.10
- Server OS: macOS Tahoe 26.5.2
- Connection: Local network, remote via domain

### Screenshots or Videos

Image

```shell

```

### Additional information

It's the same component as in the library, there it probably shouldn't display the show name as we're already in the show.

amirlanesman:
# 📦 Pull Request

## 📝 Description
Change the episode title and subtitle on the home page to display the tv show name.
This kind of display is more consistent with default web jellyfin client and more usable in case posters etc aren't good enough.
The caveat is loosing the episode duration text which is a lose but we can later overlay it on the episode thumbnail the same as other clients do.

## 🏷️ Ticket / Issue
Fixes #1886 

Before:
image

After:
image

## ✅ Checklist

- [*] I’ve read the [contribution guidelines](CONTRIBUTING.md)
- [*] Verified that changes behave as expected for all platforms
- [*] Code passes lint/formatting and type checks (`tsc`/`biome`)
- [*] No secrets, hardcoded credentials, or private config files are included
- [*] I've declared if AI was used to assist with this PR (by uncommenting the line at the bottom, or not)

## 🔍 Testing Instructions

1. Open the app on TV
2. After the hero (if exists) all episodes should now show the tv show name.
3. Go into a series.
4. Check that the episodes still look correct and don't display the show name.

## Summary by CodeRabbit

* **New Features**
  * TV episode posters now display the show name alongside episode details.
  * Episode posters show the episode title as supporting text when the show name is displayed.
  * Updated poster layouts improve title, subtitle, and episode information readability in applicable collections.


fredrikburmester (approved):

Gauvino (dismissed):
Didn't test it but code looks good

components/tv/TVPosterCard.tsx
```diff
@@ -258,11 +262,18 @@ export const TVPosterCard: React.FC<TVPosterCardProps> = ({
           )}
           {duration && (
             <>
-              <Text style={{ color: "#FFFFFF", fontSize: typography.callout }}>
+              <Text style={{ fontSize: typography.callout, color: textColor }}>
                 •
               </Text>
-              <Text style={{ fontSize: typography.callout, color: "#FFFFFF" }}>
-                {duration}
+              <Text
```
fredrikburmester, line 279: This shouldn't be inside of the duration block. Even it items should never have no duration, it's semantically incorrect. Please fix!

fredrikburmester (dismissed):

Gauvino (dismissed):

components/tv/TVPosterCard.tsx
Gauvino: Fixed in 2b1334f8. The episode name no longer sits inside the duration block, it is computed up front and the row gates on the value it actually renders, so an item without RunTimeTicks keeps its name.

components/tv/TVPosterCard.tsx
```diff
@@ -236,6 +239,11 @@ export const TVPosterCard: React.FC<TVPosterCardProps> = ({
       const duration = item.RunTimeTicks
         ? runtimeTicksToMinutes(item.RunTimeTicks)
         : null;
+      const textColor = displayShowName ? "#9CA3AF" : "#FFFFFF";
+      // When the show name is the title, the episode name takes the slot the
+      // duration usually occupies. Gate on what is actually rendered, so an
+      // item without RunTimeTicks does not silently drop its episode name.
+      const trailingText = displayShowName ? item.Name : duration;
```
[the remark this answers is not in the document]
Gauvino: Right, that duplication was on me. Fixed in 8a85b526: the trailing copy is dropped when SeriesName is missing, and an empty SeriesName is treated as missing.

No regression test: this repo has no component test harness since the jest rig was removed in #1833, and the tests here are pure-function only.

Gauvino (approved):

Gauvino:
On the outside-diff note about TVTypography: that one does not apply.

`components/tv/TVTypography` does not exist. TVTypography lives in `constants/TVTypography.ts` and is a constants object plus a `useScaledTVTypography()` hook, not a component, so there is nothing to swap the Text nodes for. `TVPosterCard` already calls that hook and applies `typography.callout` to all 12 of its Text nodes, including the ones this PR touches.

The guideline it cites points at that non-existent path in CLAUDE.md, which is what I will fix separately.

PR #1890, opened 2026-07-28
fix(player): stop reporting playback the server cannot trust

Gauvino:
# 📦 Pull Request

## 📝 Description

Four fixes in the same code path, one per commit, found while checking on device what the server actually receives after #1872, #1879 and #1873. Every claim below was measured against `LastPlaybackCheckIn` on the session, not read off the code: the server extrapolates `PositionTicks` between reports, so a moving position in `/Sessions` proves nothing on its own.

### 1. A session that keeps playing after you leave the player

**What you see today.** You exit the player, and the Jellyfin dashboard still shows you watching that episode, with the position climbing on its own. It stays that way until the server gives up on the session. Anyone sharing the server sees it, playback statistics count time nobody watched, and anything driven by session state acts on a session that ended minutes ago.

**Why.** Leaving reports "stopped", but the MPV view still holds the `onProgress` callback from the render before the exit, and destroying the player emits one last paused event that lands in the transition effect. Either one reports progress right after the stop, and a progress report for an item the server considers stopped puts it back in the session. From there the server extrapolates the position, which is why the counter keeps moving with no client sending anything.

**Fix.** `isPlaybackStopped` gets a ref mirror set synchronously by `stop()`, and both report paths read the ref instead of the state. A report already on its way now bails out. Same shape as the `isPlayingRef` mirror added in #1879, for the same class of race.

### 2. One request per second, and a full database rewrite per second offline

**What you see today.** Nothing directly, but the player sends its position to the server on every MPV tick, which is once a second: about 1300 requests over a 22 minute episode. On a downloaded item it is worse. Each report also rewrites the entire downloads database to storage and invalidates two query keys, so with a large downloads library that is a full serialize and write every second during playback.

**Why.** `onProgress` reported unconditionally. The heartbeat and the state changes were the same code path.

**Fix.** Report right after a seek, since the position jumped, and otherwise every ten seconds. State changes are unaffected: pause, resume, track and mute changes already report on their own through the effect that runs after the state commits, so they stay immediate. Measured: reports land exactly ten seconds apart, and pause and resume still reach the server in under a second.

### 3. A downloaded item was never a real session for the server

**What you see today.** Watch a downloaded episode with the server reachable and it does not show up in your playback history or statistics, even though the app is talking to the server the whole time. The session also claims `DirectStream` while the file is local.

**Why.** #1873 made the stop report connectivity-based, which is what closes the session, but the start report was still gated on the `offline` route flag. The server received a stop for a playback it never saw begin: it logs `Playback stop did not have a tracker` and drops the entry.

**Fix.** Gate the start report on connectivity too, so a download played online opens the session it later closes, and report `DirectPlay` for local playback. Connectivity is read through a ref rather than a dependency: as a dependency, a network flap mid-playback would fire a second `PlaybackStart` and move the session back to the position playback began at.

### 4. Stop position on an in-place episode switch

Hardening, not an observed failure. The stopped report is keyed on the loaded item, while the effect that prepares the incoming episode resets the shared position as soon as the route param changes. Both currently land in the same commit, so the report runs first, but only because the media source id happens to change along with the item. If it ever did not, the report would carry a position of zero and the episode you just left would lose its resume point. Depending on the route param as well makes the order explicit.

## 🏷️ Ticket / Issue

N/A. Follows #1872, #1879 and #1873.

N/A, no UI change. The effects are visible in the Jellyfin dashboard under Dashboard → Active Devices.

## ✅ Checklist

## 🔍 Testing Instructions

Open **Dashboard → Active Devices** in the Jellyfin web UI alongside the app. Note that the position shown there keeps moving between reports on its own, so judge the reports themselves, not the counter.

**No session left behind**
1. Play any item for about twenty seconds, then exit the player
2. Confirm the device disappears from Active Devices and stays gone
3. Repeat immediately: play, exit, play, exit. The race is timing dependent, so a single pass can miss it

**Report cadence**
1. Play a streamed item and leave it running for a minute
2. Confirm the session stays alive and the reported state stays correct
3. Pause, wait ten seconds, resume, several times in a row
4. Confirm the dashboard flips to paused and back within about a second each time

**Downloaded item with the server reachable**
1. Play a downloaded item from the Downloads screen with network on
2. Confirm the dashboard shows it playing, and the play method is direct play
3. Exit, and confirm the item appears in the user's playback history

**In-place episode switch**
1. With two consecutive downloaded episodes, play the first, use the next episode button, then the previous one, then exit
2. Confirm each segment is reported at its own position and none is reported at zero

**Offline**
1. Enable airplane mode, play a downloaded item, exit
2. Confirm playback works, no crash, and the local resume position advanced

### Results measured while writing this

Android emulator, release build, server reports counted from `LastPlaybackCheckIn` and the server log.

| Case | Result |
|---|---|
| Streamed, one minute | reports exactly 10 s apart, pause and resume reported in under a second |
| Exit, twice in a row | one final report, session idle and stays idle |
| Same case before the fix | session came back 4 s after exit and kept extrapolating for about 90 s |
| Downloaded, network on | play method direct play, start tracker created, stop matched to it |
| Downloaded, before the fix | `Playback stop did not have a tracker`, playback dropped from history |
| In-place switch E1 → E2 → E1 | three reports, positions 752418 ms, 43753 ms, 810059 ms, none at zero |
| Airplane mode | no report attempted, no crash, local resume advanced |

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved playback start/stop and progress reporting reliability during stop, in-place episode switches, and exiting the player.
  * Prevented stale progress and URL position updates from being sent after playback ends.
  * Refined playback-mode detection for consistent controls and reporting across streaming and downloaded scenarios.
* **Performance**
  * Throttled server playback progress reporting to reduce update frequency, while keeping immediate position updates after seeks and periodic refreshes.

### DIFF
--- a/app/(auth)/player/direct-player.tsx
+++ b/app/(auth)/player/direct-player.tsx
@@ -239,6 +239,16 @@ export default function DirectPlayerPage() {
   const { nextItem, previousItem } = playbackManager;
   const { isConnected } = useNetworkStatus();
 
+  // usePlaybackManager returns a new object on every render, so the reporters
+  // below cannot list it as a dependency without being rebuilt constantly.
+  // Mirror the progress reporter instead: it closes over connectivity and the
+  // api, and a handler captured while either was different would keep
+  // reporting through the stale one for the rest of playback.
+  const reportProgressRef = useRef(playbackManager.reportPlaybackProgress);
+  useEffect(() => {
+    reportProgressRef.current = playbackManager.reportPlaybackProgress;
+  });
+
   // Resolve audio index: use URL param if provided, otherwise use stored index for offline playback
   const audioIndex = useMemo(() => {
     if (audioIndexFromUrl !== undefined) {
@@ -734,7 +744,7 @@ export default function DirectPlayerPage() {
     // instead of casting so a gap between the item and its stream can't reject.
     const progressInfo = currentPlayStateInfo();
     if (!progressInfo) return;
-    void playbackManager.reportPlaybackProgress(progressInfo);
+    void reportProgressRef.current(progressInfo);
     // isPlaying is what makes a transition land here: currentPlayStateInfo now
     // reads the play state through a ref, so its identity no longer changes
     // when the user pauses or resumes.
@@ -819,7 +829,7 @@ export default function DirectPlayerPage() {
 
       const progressInfo = currentPlayStateInfo();
       if (!progressInfo) return;
-      playbackManager.reportPlaybackProgress(progressInfo);
+      void reportProgressRef.current(progressInfo);
     },
     [
       // Depend on the builder itself rather than re-listing what it reads: it
--- a/app/(auth)/player/direct-player.tsx
+++ b/app/(auth)/player/direct-player.tsx
@@ -376,6 +376,15 @@ export default function DirectPlayerPage() {
     isConnectedRef.current = isConnected;
   }, [isConnected]);
 
+  // The item the player is on as of the last commit. The MPV view holds the
+  // callbacks it was last rendered with, and on an in-place switch those close
+  // over the outgoing item for both the loaded item and the route param, so
+  // they agree with each other and pass a comparison between the two.
+  const currentItemIdRef = useRef(itemId);
+  useEffect(() => {
+    currentItemIdRef.current = itemId;
+  }, [itemId]);
+
   const releaseLiveStream = useCallback(
     (liveStreamId: string | null) => {
       if (!liveStreamId || !apiRef.current || offline) return;
@@ -659,9 +668,11 @@ export default function DirectPlayerPage() {
 
     // A downloaded item plays from a local file: it is neither streamed nor
     // transcoded, so report DirectPlay rather than mislabelling the session.
-    let playMethod: PlaybackProgressInfo["PlayMethod"] = "DirectStream";
-    if (offline) playMethod = "DirectPlay";
-    else if (stream.url.includes("m3u8")) playMethod = "Transcode";
+    const playMethod: PlaybackProgressInfo["PlayMethod"] = offline
+      ? "DirectPlay"
+      : stream.url.includes("m3u8")
+        ? "Transcode"
+        : "DirectStream";
 
     return {
       ItemId: item.Id,
@@ -743,8 +754,11 @@ export default function DirectPlayerPage() {
       // An in-place episode switch clears the stream and resets the position
       // while MPV can still emit one last tick for the previous episode.
       // Bail before touching shared state: writing that position into the URL
-      // would make it the next episode's start position.
+      // would make it the next episode's start position, and reporting it
+      // after the switch already reported "stopped" would put the outgoing
+      // episode back in the server session.
       if (!item?.Id || item.Id !== itemId || !stream) return;
+      if (item.Id !== currentItemIdRef.current) return;
 
       const { position, cacheSeconds } = data.nativeEvent;
       // MPV reports position in seconds, convert to ms
--- a/app/(auth)/player/direct-player.tsx
+++ b/app/(auth)/player/direct-player.tsx
@@ -82,6 +82,15 @@ export default function DirectPlayerPage() {
   const { width: screenWidth, height: screenHeight } = useWindowDimensions();
 
   const [isPlaybackStopped, setIsPlaybackStopped] = useState(false);
+  // Mirror of isPlaybackStopped that is already set when a report is built.
+  // The MPV view keeps the onProgress it was last rendered with, so a tick
+  // already in flight when the user leaves would still report progress after
+  // the "stopped" report and bring the session back on the server.
+  const isPlaybackStoppedRef = useRef(false);
+  const markPlaybackStopped = useCallback(() => {
+    isPlaybackStoppedRef.current = true;
+    setIsPlaybackStopped(true);
+  }, []);
   const [showControls, _setShowControls] = useState(true);
   const [isPipMode, setIsPipMode] = useState(false);
   const [aspectRatio] = useState<"default" | "16:9" | "4:3" | "1:1" | "21:9">(
@@ -570,7 +579,7 @@ export default function DirectPlayerPage() {
       playbackPosition: msToTicks(progress.get()).toString(),
     });
     reportPlaybackStopped();
-    setIsPlaybackStopped(true);
+    markPlaybackStopped();
     // Synchronously destroy the mpv instance + decoder + surface buffers
     // BEFORE the screen unmounts. Otherwise the next screen (or the next
     // episode's player) mounts while the old 4K decoder is still alive,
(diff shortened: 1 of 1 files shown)


app/(auth)/player/direct-player.tsx
```diff
@@ -604,11 +631,15 @@ export default function DirectPlayerPage() {
...
+  // itemId (the route param) is listed alongside the loaded item so the report
+  // runs in the same commit as an in-place switch, before the effect that
+  // resets the shared position for the incoming item: keyed on item.Id alone,
+  // the report could land a commit later and carry a position of zero.
   useEffect(() => {
     return () => {
       reportPlaybackStoppedRef.current();
     };
-  }, [item?.Id, stream?.sessionId, mediaSourceId]);
+  }, [itemId, item?.Id, stream?.sessionId, mediaSourceId]);
```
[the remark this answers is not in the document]
Gauvino: Confirmed, and it is the same class as the exit race this PR already covers. The callback the MPV view holds closes over the outgoing item for both the loaded item and the route param, so the two agree with each other and the existing comparison lets the tick through.

Fixed in 4f16176d by comparing against the item the player is on as of the last commit, which a stale callback cannot satisfy. The cleanup itself is left reporting stopped only: playback of the incoming item is not stopped, so marking the whole player as stopped there would suppress its own reports.

app/(auth)/player/direct-player.tsx
```diff
@@ -473,7 +490,11 @@ export default function DirectPlayerPage() {
   ]);
 
   useEffect(() => {
-    if (!stream || !api || offline) return;
+    // Gate on connectivity rather than on the offline flag: a downloaded item
+    // played with the server reachable still reports progress and, since the
+    // stop report is connectivity-based too, would otherwise close a session
+    // it never opened, which the server's playback history then discards.
+    if (!stream || !api || !isConnectedRef.current) return;
```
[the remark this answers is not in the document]
Gauvino: Not taking this one, on purpose.

The start report carries the position playback began at. Replaying it on a later online transition would move the server session back to that position until the next progress report corrects it, which is exactly why connectivity is read through a ref here rather than being a dependency.

Nothing regresses in that case either: progress reports still open the session, the same way develop does today for a downloaded item played online. What is missing is only the entry in the server's playback history, and the clean way to get it is a start report carrying the current position instead of the initial one. That is an addition rather than a fix for this PR.

app/(auth)/player/direct-player.tsx
```diff
@@ -626,6 +657,12 @@ export default function DirectPlayerPage() {
     | undefined => {
     if (!stream || !item?.Id) return;
 
+    // A downloaded item plays from a local file: it is neither streamed nor
+    // transcoded, so report DirectPlay rather than mislabelling the session.
+    let playMethod: PlaybackProgressInfo["PlayMethod"] = "DirectStream";
+    if (offline) playMethod = "DirectPlay";
+    else if (stream.url.includes("m3u8")) playMethod = "Transcode";
+
```
Gauvino: Done in 4f16176d.

app/(auth)/player/direct-player.tsx
```diff
@@ -626,6 +666,14 @@ export default function DirectPlayerPage() {
...
     if (!stream || !item?.Id) return;
 
+    // A downloaded item plays from a local file: it is neither streamed nor
+    // transcoded, so report DirectPlay rather than mislabelling the session.
+    const playMethod: PlaybackProgressInfo["PlayMethod"] = offline
+      ? "DirectPlay"
+      : stream.url.includes("m3u8")
+        ? "Transcode"
+        : "DirectStream";
+
```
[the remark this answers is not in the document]
Gauvino: Good catch, fixed in 88265d5d.

The two classifiers are now one function used by both the reports and the technical info overlay: downloaded first, then a transcoding url as transcode whatever the container, the stream endpoint as direct stream, and anything else as direct play. Checked on device that the two cases this PR measures are unchanged: a streamed episode still reports DirectStream, a downloaded one DirectPlay. What changes is the progressive transcode and the remote source, which used to be reported as direct streams.

Gauvino:
On the three outside-diff comments from the last review:

**Keep the MPV progress reporter up to date when connectivity or the api changes** — real, fixed in 07e95eb5. The handler cannot depend on the playback manager, which returns a new object every render, so it captured the reporter from the render it was built in, and that reporter closes over connectivity and the api. Playback started while the server was unreachable kept reporting through the offline path for the rest of the item. The reporter is now mirrored in a ref that both report paths read.

**Guard state-commit reports by playback generation** — not taking it. The state-commit effect reads the item and the stream from the render it runs in, not from a captured closure, so a late paused event after an in-place switch either still sees the outgoing item, in which case its report precedes the stop report that the route change fires, or already sees the cleared stream and returns early. The stale-callback case is the MPV handler, and that one is now scoped to the current item in 4f16176d.

**Serialize playback reports with teardown** — not in this PR. The ordering hole is real in principle: a progress request issued shortly before the stop can be processed after it. What this PR fixes is the app issuing reports after the stop, which is what was actually observable, a session that came back and kept extrapolating for about ninety seconds after leaving the player. Serializing every report through a queue in the teardown path is a much larger change, it delays the stop report, and I have no way to show it improves anything measurable here. Worth doing on its own if a revived session is ever observed again with these fixes in place.

app/(auth)/player/direct-player.tsx
```diff
@@ -70,6 +70,30 @@ import { writeToLog } from "@/utils/log";
...
+ * How the session is actually being played. Shared by the reports sent to the
+ * server and the technical info overlay, so the dashboard and the app never
+ * disagree about the same playback.
+ */
+function getPlayMethod(
+  source: { url?: string | null; mediaSource?: MediaSourceInfo | null },
+  offline: boolean,
+): PlayMethod {
+  // A downloaded item plays from a local file, with no server in the path.
+  if (offline) return "DirectPlay";
```
lancechant, line 85: Might be a bit wrong if you downloaded a file at a lower res, as it's technically transcoded and not the direct file😅

lancechant (approved):

app/(auth)/player/direct-player.tsx
Alexk2309: I would argue it is direct playing in this scenario because the server isn't transcoding it lol

Gauvino: PlayMethod describes the current session, not how the file was produced. The server does no work here, so reporting Transcode would show a phantom transcoding session in the dashboard.

The early return also guards a real problem: a download made with a bitrate cap stores the playbackInfo MediaSource, which still carries a TranscodingUrl. Without it, the next branch would match that stale URL and report a server transcode that isn't running.

If the concern is that the downloaded file isn't the original quality, that belongs in the technical info overlay (resolution/bitrate of the local file), not in PlayMethod.

Resolve the comment if you find that I responded to your question!

lancechant: > I would argue it is direct playing in this scenario because the server isn't transcoding it lol

Lol yeah true, just a technicality🤣

PR #1891, opened 2026-07-28
fix(player): drive the next episode countdown from the remaining time

Gauvino:
# 📦 Pull Request

## 📝 Description

On top of #1877, targeting its branch so it can go in with it.

Splitting the two triggers is right and this keeps it. What this changes is the countdown itself, because the second commit there ("pause auto-skip when playback is paused") stops an episode watched to the end from ever advancing on Android, and the original "skips too soon" is still reachable at any speed other than 1x.

### What breaks today

**An episode watched to the end never advances.** Android configures mpv with `keep-open=always` (`MPVLayerRenderer.kt:253`), so at the end of a file the player pauses itself instead of stopping. `autoAdvanceNextEpisode` is gated on `isPlaying`, and the button reacts to losing that by cancelling the animation and setting `progress.value = 0`. The fill empties a moment before it would have completed, and nothing advances.

Reproduced on the branch as it stands, Android emulator, starting Avatar S1E3 25 seconds before the end:

| One second left | End of the episode |
|---|---|
| countdown almost full | fill emptied, still on episode 3 |

The second shot is eight seconds after the end: play button showing (mpv paused itself), fill empty, still episode 3. It stays there.

**Pausing on purpose has the same effect.** The fill is emptied rather than held, so resuming starts a fresh ten second countdown with less than ten seconds of media left. It can no longer complete before the end.

**Skipping too soon is still reachable.** The countdown is a ten second wall-clock timer, started when ten seconds of *media* remain. The two only agree at 1x. At 0.75x those ten seconds of media take 13.3 seconds, so the timer completes about three seconds before the episode ends and skips over the tail. Above 1x it completes after the end.

### What this changes

The fill is derived from the remaining time rather than run as a timer of its own: `1 - remainingMs / 10000`, animated toward each new position sample so it still moves smoothly. Advancing happens when the item ends, which is either the position reaching the end or the player pausing itself inside the last position sample.

That is one source of truth instead of two clocks that have to be kept in sync:

- playback speed is already baked into the remaining time, so no early or late skip at any speed
- pausing stops the remaining time from moving, so the fill freezes where it is and nothing has to detect the pause
- seeking backwards moves it back, forwards moves it forward
- the end of the file is the end of the countdown, by definition

The countdown is also scoped to the item it counts for. Right after an in-place episode switch the remaining time still describes the previous episode, which reads as "one second left" for the new one. Without that guard the first version of this patch chained from episode 3 to episode 6 in one go, which is a good illustration of why the trigger has to be tied to the item.

### After

| Same scenario, end of the episode | Pause during the countdown |
|---|---|
| episode 4 playing | fill frozen mid way |

Episode 3 ends, episode 4 starts on its own, once. Pausing with four seconds left holds the fill where it is for as long as the pause lasts, and resuming advances at the end.

At 0.75x, where the old timer completed about three seconds early, the fill now tracks the media instead of the wall clock: one second of media left is a fill that is nearly complete, and the episode is still playing.

| 0.75x, one second of media left | Same run, after the end |
|---|---|
| still playing, fill nearly complete | episode 4 playing |

## 🏷️ Ticket / Issue

N/A. Builds on #1877.

Above. No layout change, the button is the same, only what drives its fill.

## ✅ Checklist

## 🔍 Testing Instructions

Autoplay has to be enabled, and the consecutive episode counter has to be below the limit, otherwise the "Are you still watching?" overlay takes over.

**An episode watched to the end**
1. Start an episode that has a next one, and let it reach the end
2. Confirm the next episode starts on its own, once

**Pause during the countdown**
1. In the last ten seconds, pause
2. Confirm the fill stops where it is instead of emptying, and nothing advances while paused
3. Resume, and confirm it advances at the end

**Playback speed**
1. Set the speed to 0.75x and let an episode reach the end
2. Confirm nothing is skipped before the end, and that it advances when the episode is actually over

**Credits trigger**
1. On an episode whose credits segment reports content after them, confirm the button still appears without a fill and only advances when tapped

Measured on the Android emulator, starting an episode 25 seconds before its end:

| Case | Before | After |
|---|---|---|
| Episode watched to the end | fill emptied at the end, stayed on the same episode | next episode starts, once |
| Pause with four seconds left | fill emptied, could no longer complete | fill frozen where it was, nothing advances |
| Resume after that pause | ten second countdown restarted, never completed | advances at the end |
| 0.75x speed | timer completed about three seconds before the end | still playing one second from the end, advances when the episode is over |

The autoplay episode limit was set to disabled while measuring, so the "Are you still watching?" overlay could not take over the runs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic next-episode playback to follow the current video’s actual remaining time.
  * Updated the next-episode countdown to pause, reset, and complete accurately when playback state or the video changes.
  * Improved skip and next-episode controls for paused playback and videos reaching their end.

### DIFF
--- a/components/video-player/controls/Controls.tsx
+++ b/components/video-player/controls/Controls.tsx
@@ -386,7 +386,10 @@ export const Controls: FC<Props> = ({
 
   const showNextEpisode =
     showNextEpisodeFromCredits || showNextEpisodeFromRemainingTime;
-  const autoAdvanceNextEpisode = showNextEpisodeFromRemainingTime && isPlaying;
+  // Whether reaching the end of the item may advance on its own. Pausing is
+  // not a reason to take that away: the countdown follows the remaining time,
+  // which stops moving while playback does.
+  const autoAdvanceNextEpisode = showNextEpisodeFromRemainingTime;
 
   const goToItemCommon = useCallback(
     (item: BaseItemDto) => {
@@ -647,6 +650,9 @@ export const Controls: FC<Props> = ({
             willShowNextEpisode={willShowNextEpisode}
             showNextEpisode={showNextEpisode}
             autoAdvanceNextEpisode={autoAdvanceNextEpisode}
+            remainingTime={remainingTime}
+            isPlaying={isPlaying}
+            itemId={item.Id}
             skipIntro={skipIntro}
             skipCredit={skipCredit}
             onNextEpisodeFinish={handleNextEpisodeAutoPlay}
--- a/components/video-player/controls/Controls.tsx
+++ b/components/video-player/controls/Controls.tsx
@@ -382,7 +382,8 @@ export const Controls: FC<Props> = ({
   // Driven by actual playback position vs. duration, independent of segment
   // metadata, so it's safe to auto-advance from this trigger.
   const showNextEpisodeFromRemainingTime =
-    willShowNextEpisode && remainingTime < 10000;
+    willShowNextEpisode &&
+    remainingTime < CONTROLS_CONSTANTS.NEXT_EPISODE_COUNTDOWN_MS;
 
   const showNextEpisode =
     showNextEpisodeFromCredits || showNextEpisodeFromRemainingTime;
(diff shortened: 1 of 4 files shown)


Gauvino:
@coderabbitai full review

Simon-Eklundh:
While I didn't personally see the error you're referring to (it skipped after episode had ended even after I had paused in the last 10 seconds), I am a fan of the playback timer being paused, and I did forget people watch at non-standard speeds

PR #1909, opened 2026-08-02
feat(desktop): add an Electron desktop client built from the same source

Arkniem:
# 📦 Pull Request

Code written by me
Ai used for comments and documentation

## 📝 Description

Adds a desktop build of Streamyfin for Windows and Linux, running the existing app rather than a reimplementation of it.

The renderer is the unmodified Expo web bundle, so **mobile and desktop share one source tree** — a change to a screen lands on both without being ported. Only the pieces that genuinely cannot exist in a browser are swapped out, through Metro's platform resolution:

- **Playback** — libmpv has no browser build, so `modules/mpv-player/index.web.tsx` drives an HTML5 `<video>` with hls.js attached when Jellyfin returns an HLS manifest. Limits that cannot be met (embedded track enumeration, ASS/SSA styling) return empty rather than faking success.
- **Native-only modules** — track-player, Chromecast, device-info, glass effects and secure-store get web shims under `web-shims/`.
- **Secure storage** — `expo-secure-store` is native-only, so saving an account failed outright. Electron's `safeStorage` backs it over a narrow key/value IPC channel.
- **Session persistence** — the loopback server binds a fixed port, because a random one changes the origin and silently empties `localStorage`; state is also mirrored to disk so a session survives an app update.
- **Update banner** — checks GitHub releases and offers an update at the top of the app.

Electron adds only what the page cannot do for itself. `contextIsolation` is on, `nodeIntegration` off, sandbox on, remote navigation blocked, and external links open in the real browser.

Later commits fix desktop-specific playback bugs found while testing:

- **Pause rewound and resumed by itself.** The screen passes fresh inline callbacks each render and the source effect depended on their identity, so any re-render tore the stream down, re-seeked to the start position and called `play()`. Pausing causes a re-render. The effect now keys off the source's contents.
- **Hidden controls could not be brought back.** The overlay only listened for touch events, which a mouse never fires. Clicking now toggles, and mouse movement reveals them via a listener that only ever shows, so a moving cursor cannot flicker them.
- **Fullscreen** is automatic while the player is open rather than a button, and steps aside for picture-in-picture, where the video has its own window.
- **Dropdowns did nothing.** `PlatformDropdown` only acted on a caller-controlled `open` flag on Android, so on web the trigger swallowed the press and nothing opened — the season picker among others.

## 🏷️ Ticket / Issue

N/A — no existing issue.

The app running on Windows, same UI as mobile:

Streamyfin desktop home screen

Player in fullscreen:

Desktop player in fullscreen

Picture-in-picture — the window drops out of fullscreen and shows a centred message, because Chromium's own placeholder is pinned to the top-left corner and cannot be restyled:

Desktop player in picture-in-picture

## ✅ Checklist

- [ ] Verified that changes behave as expected for all platforms — exercised end to end on the Windows desktop build; **the Linux artifacts have not been run by anyone**, and iOS/Android/TV are unaffected by design but have not been re-tested

## 🔍 Testing Instructions

1. `bun run desktop` — builds the web bundle and launches Electron
2. Connect to a Jellyfin server and log in; confirm saving an account succeeds (this is what `safeStorage` fixes)
3. Quit and relaunch; confirm the server URL and session are still there and no re-login is needed
4. Play a movie; confirm it starts, and that **pause holds** — the position must not jump back and playback must not resume on its own
5. Click the video to hide the controls, then click again and move the mouse; confirm they come back both ways
6. Confirm the window enters fullscreen when the player opens and returns to windowed when you leave it
7. Enter picture-in-picture; confirm the window leaves fullscreen, the mini player shows video rather than a blank frame, and the page shows a centred message
8. Open a series and click the season selector; confirm the season list opens
9. Check the app icon and fonts render (not tofu boxes), which exercises the bundled assets
10. For a release build: `bun run desktop:dist:win` or `desktop:dist:linux`

## Summary by CodeRabbit

* **New Features**
  * Added web and desktop app support with browser-friendly navigation and platform integrations.
  * Introduced Electron desktop builds for Windows and Linux, including secure storage, fullscreen playback, and update notifications.
  * Added web video playback with HLS, subtitles, picture-in-picture, and fullscreen controls.
  * Added desktop packaging and automated release artifacts.
* **Bug Fixes**
  * Improved compatibility by preventing unsupported mobile and TV features from affecting web and desktop experiences.
* **Documentation**
  * Added desktop setup, packaging, security, and known-limitations documentation.


Alexk2309:
My main concern with this PR is that desktop support isn't really just a matter of enabling another platform. Even though the goal is simply to support desktop, this PR introduces a significant number of changes, and I don't think that level of added complexity is worth it for a platform we're not currently targeting.

The app was originally designed with mobile in mind, so the UI doesn't naturally translate to desktop. We've already put a lot of work into bringing it to tvOS, and even that required a lot of custom UI and platform-specific changes. I think desktop would end up being a similar situation, which means another platform we'd need to maintain over time.

I'd also like to hear what the other Streamyfin developers think before we decide to support another platform. Personally, I'd rather we focus on delivering a polished experience on a smaller number of devices than trying to support everything. Moonfin has chosen to support a wide range of platforms, which is a perfectly valid approach, but I think it also comes with a much larger maintenance burden. For Streamyfin, I think it makes more sense to focus on doing a few platforms really well.

lancechant:
> My main concern with this PR is that desktop support isn't really just a matter of enabling another platform. Even though the goal is simply to support desktop, this PR introduces a significant number of changes, and I don't think that level of added complexity is worth it for a platform we're not currently targeting.
> 
> The app was originally designed with mobile in mind, so the UI doesn't naturally translate to desktop. We've already put a lot of work into bringing it to tvOS, and even that required a lot of custom UI and platform-specific changes. I think desktop would end up being a similar situation, which means another platform we'd need to maintain over time.
> 
> I'd also like to hear what the other Streamyfin developers think before we decide to support another platform. Personally, I'd rather we focus on delivering a polished experience on a smaller number of devices than trying to support everything. Moonfin has chosen to support a wide range of platforms, which is a perfectly valid approach, but I think it also comes with a much larger maintenance burden. For Streamyfin, I think it makes more sense to focus on doing a few platforms really well.

I agree with this, maybe once TVos side is polished we then come back to adding other platforms, since TV side is new and fixing a few things still especially compatibility with displays etc, so a nice to have for the future as those can the also be used too. 

@fredrikburmester your thoughts?

Arkniem:
I created this in mind so that when anything is fixed on the mobile version, it ships automatically to the desktop version that way support for another version is not required as they run on the exact same source code and builds the set up file from the repo.

It would not require any additional maintenance

Gauvino:
Im thinking of something, if in the feature we want more support on platform like desktop and else, we can create a base of code of streamyfin so the other client can use and that would not require more maintenance that we had that would be cool

PR #1919, opened 2026-08-05
fix: toggles not working in settings

lancechant:
# 📦 Pull Request

Was used to help verify the findings
[AI Assisted](#)

## 📝 Description

fixing an issue where certain toggles were not working. The settings toggles for example the set subtitle track from previous item and audio couldn't be enabled/disabled in app

## ✅ Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved settings persistence so explicitly updated values are retained, including values matching application defaults.
  * Ensured plugin settings consistently reflect meaningful selected values.
  * Updated settings immediately when changes are explicitly saved, even if the value remains unchanged.
  * Prevented saved preferences from being unintentionally replaced by default values during runtime settings resolution.


Gauvino:
Can you explain more in the description waht problem you encounter ?

lancechant:
> Can you explain more in the description waht problem you encounter ?

We discussed it on discord.. but it's the settings toggles for example the set subtitle track from previous item and audio couldn't be enabled/disabled in app

Gauvino:
Superseded by #1944, which moved this policy into `utils/atoms/settingsOverrides.ts` and fixes the same root cause: an unlocked plugin value was pinned at read time, so pressing a toggle wrote the value already in storage and the write was dropped as a no-op. It ships unit tests covering exactly this regression (`an unlocked plugin value never overrides the stored value`, on `rememberAudioSelections`).

PR #1920, opened 2026-08-06
refactor: consolidate delete confirmations into one hook

fredrikburmester:
# 📦 Pull Request

[AI Assisted](#)

## 📝 Description

Consolidate every delete confirmation behind one `useConfirmDelete` hook. They were all implemented separately, and `SeriesCard` left out `cancelButtonIndex`, so on iOS its Cancel button rendered inline under Delete instead of detached at the bottom like the movie one. Deleting a series and deleting a movie looked like two different dialogs.

The hook shows an action sheet on mobile and an Alert on TV (a remote can't navigate an action sheet). The download cards, storage settings and the playlist sheet all use it now. Delete/Cancel labels come from i18n instead of hardcoded English, so they were never translated before. No new translation keys.

Three other things while I was in there:

- The bulk delete buttons on the downloads page now confirm. "Delete all" wiped every download on a single tap.
- Dropped the outer `TouchableItemRouter` around `MovieCard` on the downloads page. The card renders its own, so two long press sheets were stacked on one card and only the inner one was reachable.
- Removed `EpisodeCard.tsx`, nothing imported it.

Note on the last one: I checked the responder behaviour by reading the code, not on a device. If the outer touchable was somehow winning, movies on the downloads page would lose the mark-as-played/favourite menu. Worth a look when testing.

`StorageSettings` and `PlaylistOptionsSheet` go from a native Alert to an action sheet on mobile. That's the point of the change, but it is a visible difference on Android where you now get the JS bottom sheet instead of a Material dialog.

## 🏷️ Ticket / Issue

N/A

N/A

## ✅ Checklist

**Not tested on device yet.** Typecheck, biome check and format all pass, but nothing here has been run on iOS, Android or TV.

## 🔍 Testing Instructions

1. Download a movie and a few episodes of a series.
2. Go to the downloads page. Long press the movie, then long press the series. Both sheets should have Delete on top in red and Cancel detached at the bottom.
3. The series sheet should show the series name and the episode count.
4. Confirm on each and check the download actually goes away.
5. Open the size button in the header. Tap "Delete all movies" and cancel, nothing should be deleted. Tap it again and confirm.
6. Settings > Storage > delete all downloaded files. Should be an action sheet now, not an Alert.
7. Music > a playlist > options > delete. Same sheet.
8. Repeat 2 and 6 in Swedish, the buttons should say "Ta bort" and "Avbryt".
9. On TV, storage settings should still show an Alert with Cancel first.

https://claude.ai/code/session_019xkHUWrfqjYaSK4ihHBBPf

## Summary by CodeRabbit

* **New Features**
  * Added consistent confirmation prompts before deleting downloads, playlists, and stored data.
  * Confirmation dialogs now display relevant item details, such as titles and episode counts.
  * Added localized confirm and cancel actions across supported platforms.

* **Bug Fixes**
  * Prevented accidental bulk deletion by requiring confirmation.
  * Preserved deletion feedback, error handling, and storage updates after confirmation.


lancechant (approved):

Gauvino (approved):
If this as been tested, it's cool otherwise the code look's ok

PR #1901, opened 2026-07-31
feat: add native hero media-bar on the home screen

eGirlQuint:
# 📦 Pull Request

[AI Assisted](#) 
Most of the code in this PR was copied from prexisting code in the code base and #1466.
AI was used to write some new code, merge the best parts of both implementations, and to help me find the right spots around the code base.
This is my first PR, and first time working with mobile apps. I am very open to feedback!
I am used to using gitlab, so you may have to watch me fight a bit with github.

## 📝 Description

This PR adds a hero media bar on the home screen of the mobile app.
The appropiate settings were also readded.
For more details, please read #1702 

## 🏷️ Ticket / Issue

Resolves #1702 

## Acceptance criteria

- [x] `showLargeHomeCarousel` is restored to the `Settings` type and `defaultValues` (default `false`) in `utils/atoms/settings.ts`.
- [x] A toggle bound to `t('home.settings.other.show_large_home_carousel')` appears in `AppearanceSettings.tsx`; flipping it shows/hides the hero with no app restart.
- [x] `components/home/PromotedHeroCarousel.tsx` exists and renders on **phone** at the top of `HomeMobile`, inside the existing home `ScrollView`.
- [x] Hero items come from the **shared** Continue Watching + Next Up data path (extracted hook reused by both `Home.tv.tsx` and the phone hero); deduped and capped at 15; no duplicate/divergent query.
- [x] Slides show backdrop + logo (fallback to title) + truncated overview; tapping a slide navigates to that item's detail screen.
- [x] Auto-rotates on a configurable interval; pauses on user drag and when the home screen is not focused; rotation interval is exposed as a constant/setting (not a hidden magic number).

- [ ] No hardcoded colors — theme tokens only, consistent with `TVHeroCarousel`.
I could not find any theme tokens in TVHeroCarousel. So they have not been used in this PR.

- [x] When the setting is `false`, no hero query fires and nothing renders (no wasted network/render).
- [x] The orphaned `home.settings.other.show_large_home_carousel` key now has a real reader; `git grep showLargeHomeCarousel` returns code matches again.

- [ ] TV behavior (`showTVHeroCarousel` / `TVHeroCarousel`) is unchanged.
Have to look into this: It seems a media their logo was previously excluded from the data fetch. Which means the logo handling never triggered, and always used the title as fallback. This might be an unintended bug fix. 

- [x] biome/lint clean; LF line endings.

### 🖼️ Screenshots / GIFs (if UI)

image
image
image
image
image

## ✅ Checklist

[...]

Issue #1831, opened 2026-07-20
chore: pre-release dependency refresh & Renovate cleanup (tracking)


Gauvino:
# 🎯 Goal

Keep the dependency surface healthy ahead of the next release: land every safe update in
grouped PRs, take the majors one at a time with a real device test, retire the dead and
abandoned packages, and fix the Renovate rules that keep producing bad PRs.

The Expo SDK 56 to 57 migration that opened this issue is done and shipped, so it is out of
scope now. The history is kept at the bottom for reference.

## 📌 State of play (verified 2026-08-12 against npm, GitHub releases and the published tarballs)

- **Renovate's Expo SDK group is producing a downgrade.** The now closed #1895 proposed
  `react-native 0.86.0-2` to `0.86.0-0rc3`, which is a move backwards: Renovate reads
  `react-native-tvos` prerelease versioning the wrong way round. That drags an otherwise valid
  `expo` and `react` bump down with it, so the rule needs a fix before the group is usable again.
- **`react-native-pager-view` 9.0.0 went stable on 2026-08-11, and it does not build.** Tested
  locally: `:react-native-pager-view:compileDebugKotlin` fails with
  `Unresolved reference 'viewpager2'`. The Compose rewrite
  (callstack/react-native-pager-view#1092, the only change in 9.0.0) swapped the Gradle
  dependencies to `androidx.compose.*` and dropped `androidx.viewpager2`, but
  `NestedScrollableHost.kt` and `PagerViewViewManagerImpl.kt` still import it, and both files
  ship in the published tarball. Nothing on our side: Kotlin 2.1.20 and the Compose compiler
  plugin resolved fine. There is no 9.0.1 and no upstream report yet. Hold at 8.x, take 8.0.5,
  and file it upstream. Details on #1946.
- **`react-native-gesture-handler` v3 is still not viable for us**, now with evidence rather
  than an assumption. #1610 closed, see the section below.
- **`react-native-reanimated-carousel` is dead on `develop` but not in the open PRs.** Its only
  importer here is `components/home/LargeMovieCarousel.tsx`, which nothing imports, so #1954
  deletes both. But #1901 adds `components/home/PromotedHeroCarousel.tsx`, which imports the
  package on the v4 API, so the dependency drop is contingent on the merge order and #1954 is a
  draft until that is decided. #1883 was closed on the "no live usage" basis; if #1901 lands,
  the v5 migration comes back and applies to `PromotedHeroCarousel.tsx`.

[...]

PR #1656, opened 2026-06-02
test(ui): add Maestro mobile E2E automation

jeeftor:
# 📦 Pull Request

[AI Assisted](#)

## 📝 Description

Draft stacked PR on top of #1597.

This PR adds a repeatable Maestro/Jellyfin fixture harness for validating the custom-header login work from #1597 with real app navigation, a real Jellyfin container, deterministic fixture media, uploaded CI artifacts, and Android phone/Android TV playback coverage.

### Stack Shape

- Base PR: #1597, `jeeftor:feature/jeeftor-tv-test`
- This PR branch: `jeeftor:feature/maestr-dev`
- Focused local diff: `feature/jeeftor-tv-test..feature/maestr-dev`
- Commit shape: one commit on top of the #1597 branch, `test(ui): add Maestro fixture harness`

GitHub cannot set the upstream PR base to `jeeftor:feature/jeeftor-tv-test` because that branch lives in a fork. Until #1597 merges, this draft targets upstream `develop`, so GitHub's visible diff includes the #1597 layer. Reviewers should use the stacked diff above for the focused Maestro harness changes.

#### Maestro Flows

- Adds reusable Maestro flows for:
  - Android simple login
  - Android Cloudflare/custom-header login
  - iOS simple login
  - iOS Cloudflare/custom-header login
  - Android phone authenticated fixture playback for Steamboat Willie
  - Android TV authenticated fixture playback for Steamboat Willie
  - iOS authenticated fixture playback for Steamboat Willie
- Adds shared flow fragments for launch, server connection, and login steps.
- Captures screenshots into `tests/maestro/artifacts/<timestamp>-<flow>/`.
- Records Android phone and Android TV screen videos for CI playback flows.
- Copies Maestro debug artifacts on failures when Maestro creates `$HOME/.maestro/tests` output.

#### Jellyfin Fixture

- Adds `tests/fixtures/jellyfin` with Docker Compose, a saved Jellyfin `base_config`, fixture scripts, API smoke checks, and public-domain media download/transcode helpers.
- Uses `admin` / `admin` for the saved fixture login.
- Provides deterministic library shapes for movies, shows, and music.
- Publishes the correct Jellyfin URL for Android emulator, iOS simulator, or physical-device targets into `tests/maestro/.env.local`.
- Adds selector discovery for playback flows, including fixture media IDs written into the Maestro env file.

#### Local Developer Commands

- Expands the root `Makefile` with targets for:
[...]

PR #1954, opened 2026-08-12
chore(deps): drop react-native-reanimated-carousel and its dead component

Gauvino:
# 📦 Pull Request

[AI Assisted](#)

## 📝 Description

`components/home/LargeMovieCarousel.tsx` is dead code. Nothing imports it: `git grep LargeMovieCarousel` returns only its own definition. The last reference outside the file disappeared in an old home refactor, so it has not rendered in a long time. It is also the only thing on `develop` that imports `react-native-reanimated-carousel`.

This deletes the component and drops the package.

> [!IMPORTANT]
> **The package removal conflicts with #1901.** That PR adds `components/home/PromotedHeroCarousel.tsx`, which imports `react-native-reanimated-carousel` with the v4 API (default `Carousel` export, `ICarouselInstance`, `Pagination`). So the package is dead on `develop` today but a pending feature revives it.
>
> Deleting the dead component is correct either way. Only the dependency drop is contingent, so this is a draft until the ordering is decided:
> - if #1901 lands first, drop the `package.json` and lockfile part of this PR and keep only the component deletion;
> - if this lands first, #1901 has to add the dependency back. It is already `CONFLICTING` and needs a rebase regardless.
>
> #1656 and #1597 also touch the dead file, but only with an incidental import change. A delete/modify conflict there is resolved by keeping the delete. Both are already conflicting.

On the open v5 bump: v5 is not a drop-in. It removes the default export, renames `ICarouselInstance` to `CarouselRef`, turns `onProgressChange` into a `(progress: number) => void` callback instead of accepting a Reanimated `SharedValue`, renames `containerStyle` to `contentContainerStyle` and replaces `mode`/`modeConfig` with `layout`. #1883 was closed on the basis that no live code used it. If #1901 lands, that migration work comes back and applies to `PromotedHeroCarousel.tsx`.

Left alone on purpose: the `usePopularPlugin` setting is now read by nothing, but it is still part of the admin plugin's lockable settings surface, so dropping it is a separate decision.

## 🏷️ Ticket / Issue

Part of #1831. Related to #1883 and #1901.

N/A. The component never rendered, so there is nothing to show before or after.

## ✅ Checklist

## 🔍 Testing Instructions

1. `bun i` and check the app still boots on phone and TV.
2. Open the home screen on both. Nothing should be missing, since the removed component was never mounted.

Local checks: `bun run typecheck` passes, `bun test` 138/138, `bun run i18n:check` clean, `bun install --frozen-lockfile --dry-run` clean.

Note on `bun run doctor`: it reports 16 Expo packages out of date on this branch, but that is pre-existing on develop and unrelated. It is the Expo patch drift tracked in #1831, and none of the listed packages is this one.


datwildfire:
is this about the first page banner for mobile? if so really sad

Gauvino:
Correcting myself: I checked `develop` for usage but not the open PRs.

#1901 adds `components/home/PromotedHeroCarousel.tsx`, which imports `react-native-reanimated-carousel` on the v4 API. So the package is dead on develop today, but that pending feature brings it back.

Moved to draft. Deleting the dead component holds either way, the dependency drop is the part that depends on which lands first.

PR #1998, opened 2026-08-23
feat(playback): resume in place by default

Gauvino:
# 📦 Pull Request

[AI Assisted](#)

## 📝 Description

Clicking Play on an in-progress item now resumes right where you left off, instead of first popping the "Resume playback: continue where you left off or start from the beginning?" dialog. Carrying on watching is the common case and should be one tap.

This is a single default flip: `showResumeDialog` in `defaultValues` goes `true` to `false`.

Nothing else was needed. The resume/restart branch already exists in both entry points and keys off this setting:

- Mobile: `components/PlayButton.tsx` (`if (progressTicks > 0 && !settings.showResumeDialog) resume`)
- TV: `components/ItemContent.tv.tsx` (same guard)

The prompt is not removed. The "Ask to resume or start over" toggle stays in playback settings on both platforms (`PlaybackControlsSettings.tsx` and `settings.tv.tsx`), and admins can still lock it through plugin settings, so anyone who prefers being asked just turns it back on.

## 🏷️ Ticket / Issue

N/A

No new UI. The change removes a dialog on an existing button, it does not add or restyle anything.

## ✅ Checklist

Note on the platform box: this has not been run on a device yet. Both call sites were read and the existing `!showResumeDialog` branch already handles the resume path, and `bun run typecheck` plus biome are green, but that is not a substitute for a device pass. Leaving it unticked until someone actually plays a half-watched item on phone and TV.

## 🔍 Testing Instructions

1. Start any movie or episode, watch a minute or two, then leave the player so the item carries playback progress.
2. Go back to that item and press Play. It should resume at your position straight away, with no "resume or start over" dialog.
3. Open Settings > Playback controls and turn on "Ask to resume or start over".
4. Press Play on the same in-progress item again. The dialog should be back, with "Play from start" and "Continue from ..." both working as before.
5. Repeat 2 and 4 on TV, where the same guard lives in `ItemContent.tv.tsx`.
6. Press Play on an item with no progress. It should start from the beginning either way, with no dialog.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resume prompts are now disabled by default.
  * Users will no longer see the resume dialog unless they enable it in settings.


Gauvino:
@coderabbitai review

Simon-Eklundh:
I prefer this personally, but i don't think cage agrees (and we had a whole discussion about it in #develop iirc)

Gauvino:
> I prefer this personally, but i don't think cage agrees (and we had a whole discussion about it in #develop iirc)

You mean by default?
Cause by default I prefer this behavior, and this was discussed also on Jellyfin. I think that is why I preferred to follow the Jellyfin behavior

Simon-Eklundh:
> > I prefer this personally, but i don't think cage agrees (and we had a whole discussion about it in #develop iirc)
> 
> You mean by default?
> Cause by default I prefer this behavior, and this was discussed also on Jellyfin. I think that is why I preferred to follow the Jellyfin behavior

Yeah, on by default 

I don't care much either way since it's settable in plugin

Gauvino:
> > > I prefer this personally, but i don't think cage agrees (and we had a whole discussion about it in #develop iirc)
> > 
> > 
> > You mean by default?
> > Cause by default I prefer this behavior, and this was discussed also on Jellyfin. I think that is why I preferred to follow the Jellyfin behavior
> 
> Yeah, on by default
> 
> I don't care much either way since it's settable in plugin

That's why I prefer it this way

Gauvino:
@lancechant @whoopsi-daisy your guys opinion on this ?

lancechant (approved):

lancechant:
> @lancechant @whoopsi-daisy your guys opinion on this ?

Was voted as disabled by default so we go with that

PR #2003, opened 2026-08-26
fix(player): stop reporting position 0 when exiting the native player

lancechant:
# 📦 Pull Request

[AI Assisted](#)

## 📝 Description
Exiting the native player could report `PositionTicks: 0` (server log: "Stopped at 0 ms") after a long watch session, or a stale pre-seek position. Each source is closed where it lives:

- **Android `PlayerViewModel` renderer callbacks**: `MPVLayerRenderer.stop()` zeroes its cached position, and callbacks already posted to the main handler read those fields when they run, so a tick delivered after `onDismiss` handed JS position 0 while teardown was still awaiting its reports. Every renderer callback is now dropped once `isTearingDown` is set, matching iOS; a late `onTracksReady` would otherwise run the auto-subtitle path against a closed player.
- **`onDismiss` position (`NativePlayerSession` on both platforms)**: iOS read `engine.getCurrentPosition()`, which is 0 until mpv's first time-pos callback and keeps the outgoing file's position across an in-place swap; Android read the interpolated UI clock, which a touch drag moves before any seek is issued. Both now emit the view model's authoritative position (`position` / `authoritativePosition`, now `private set`): seeded from the stream's start position, written by every authoritative tick and synchronously by every seek, never by the display clock or a scrub. A dismiss landing mid-drag reports where playback is paused, and a playing exit reports the last tick (up to 1 s behind the clock, by design).
- **Every seek goes through the view model**: iOS lock-screen, Control Center and PiP seeks now route via the engine delegate (`requestsSeekTo` / `requestsSeekBy`) and the WS remote seek via `viewModel.seek(to:)`, so they move the tracked position and arm the seek report like a scrubber seek; Android already did. The inline `MpvPlayerView` host keeps the engine-direct behaviour through protocol defaults.
- **Android `NativePlayerModule.seekTo` on the main queue**: the module function ran on Expo's background AsyncFunction queue, and `viewModel.seekTo` syncs the window's keep-screen-on flag, which the view hierarchy rejects off the main thread (`CalledFromWrongThreadException`, visible as a rejected `NativePlayer.seekTo` promise). The throw came after the mpv seek but before the tracking tick, so a backward remote seek followed by an exit reported the pre-seek position. It now runs on the main queue, as the iOS function already did.
- **Seek reporting** (`seekTo` / `seek(to:)`): the immediate `didSeek` emit added during review reported the *requested* target and disarmed the landed-tick report, so with keyframe seeks the server sat on a position playback never reached until the next interval. The immediate tick is now `trackingOnly` (moves the JS tracked position, never reported itself) and `pendingSeekReport` stays armed, so the first authoritative tick carries `didSeek` with the landed position. The segment check runs after the progress emit on both platforms, so an auto-skip's own tick follows the tick that triggered it instead of being consumed by it.
- **JS `teardownSession` (`providers/NativePlayerProvider.tsx`)** resolves the final position through `resolveFinalPositionMs` (`utils/nativePlayer/`) — the later of the tracked and native values, non-finite input rejected — and passes the same ticks into `reportPlaybackStopped`. The two inputs are equal in every reachable state, so the max is a regression floor; unit tests pin it. The `onProgress` listener body moved into `applyProgressTick` with tests for the tracking-only, seek and interval gates, and `PROGRESS_REPORT_INTERVAL` (defined in both players) now lives in `constants/Playback.ts`.

A genuine stop at 0:00 still reports 0.

## 🏷️ Ticket / Issue
N/A (no open issue found for this)

N/A

## ✅ Checklist

## 🔍 Testing Instructions
The Jellyfin server log is the ground truth for every step: watch for the `Playback stopped ... Stopped at X ms` line and the item's resume position in the web client.

1. Play a long item with the native player, let it run a few minutes, pause, wait, then exit. The stop line must carry the real position, not 0.
2. Resume a partly watched item and exit within a second of the player appearing. The resume point must be preserved.
3. Start a fresh item and stop it at 0:00. The stop line must say 0.
4. Seek forward and backward with the scrubber, then exit right after the backward seek. The stop line must carry the seek target, not the pre-seek position.
5. From the Jellyfin web client's Play On bar, seek the app's session backward, then send Stop. The stop line must carry the remote seek target.
6. After a scrubber seek, check the next progress report in the log: it should land within a second with the position mpv actually reached, not the requested one.
7. iOS: from the lock screen or Control Center, drag the position back, then close the player before the next tick. The stop line must carry the lock-screen target.
8. Android: drag the scrubber forward without releasing, and let the sleep timer (or a remote Stop) close the player mid-drag. The stop line must carry the paused position, not the thumb.
9. Repeat 1, 2 and 4 on Android, where the zeroed-tick path lives.

### Verification so far
Server-driven runs (Play / Seek / Stop over the Jellyfin session API, results read back from the server) on the iOS simulator and the Android emulator: every exit, seek and throttle scenario reported the expected position on both platforms, including 10/10 early exits at the seeded resume point and 10/10 exits racing the last tick. After the switch of `onDismiss` to the authoritative position and the seek routing changes, the exit legs were re-run on the iOS simulator against the final build (provenance checked: installed binary and Metro both from this branch): early exit 10/10 at the seeded 15:00, exit after a 90 s pause, exit racing the last tick 10/10, backward scrub then exit, remote backward seek then stop, seek report and landed position — all as expected. The same legs were then run on the Android emulator against the final build (provenance checked the same way). The first run exposed the off-main `seekTo` throw above: remote backward seek then stop stored 25:00 instead of 5:00, and logcat showed every remote seek rejecting its promise. With `seekTo` on the main queue all nine legs pass — early exit 10/10 at the seeded 15:00, exit after a 90 s pause at the paused position, exit racing the last tick 10/10 by server Stop and 5/5 by the hardware back key, remote backward seek then stop at 5:00, forward remote seek then stop, seek report throttling, landed position — with no rejected `seekTo` across 54 mpv seeks. Two Android facts the harness has to allow for: mpv runs with `hr-seek=no`, so a seeded position lands on the previous keyframe (a 10:00 seed plays from about 9:51, and the stored position after the first tick honestly says so), and Jellyfin extrapolates `PlayState.PositionTicks` by wall clock between reports, so the session position is not evidence of what the client reported. Physical devices and TV are not covered yet, hence the unticked platform box.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved the player’s last known position when native teardown reports an uninitialized position of zero.
  * Updated seeking to report progress immediately with the confirmed seek position across supported platforms.
  * Clamped seek positions to the media duration when available.
  * Improved position accuracy for remote-control, lock-screen, Control Center, and picture-in-picture seeking.
  * Prevented teardown-related events from overwriting the final playback position.
  * Standardized playback progress reporting intervals across player experiences.


lancechant:
Hey @Gauvino I see the tests are fixed now, mind taking a look when you free

Gauvino:
> Hey @Gauvino I see the tests are fixed now, mind taking a look when you free

I found a couple of issue with your pr, will try to make a proper review but it's gonna monday don't have any computer on me right now

PR #2016, opened 2026-08-30
fix(player): ignore an outgoing stream's EOF during an in-place swap

fredrikburmester:
# 📦 Pull Request

[AI Assisted](#)

## 📝 Description
Split out of #2003 (sibling bug, same listener block).

`beginSession` in `providers/NativePlayerProvider.tsx` points `sessionRef` at the incoming session with `awaitingLoad: true` before the new stream has loaded, while the outgoing stream keeps playing in mpv. If the outgoing stream reaches end of file inside that window, both native players emit `onPlaybackEnded` with the outgoing stream's position, and the `onPlaybackEnded` listener wrote it onto the incoming session, so the final progress and stop reports marked the new item as stopped where the previous one ended. Every sibling listener already drops events while `awaitingLoad` is set (`onNextEpisodeRequested` documents the same swap-EOF window); this one now does too.

The listener body moved into `applyPlaybackEnded` (`utils/nativePlayer/`) so the guard is unit tested, following `applyProgressTick` from #2003. No native code changes.

## 🏷️ Ticket / Issue
N/A — follow-up to #2003.

N/A

## ✅ Checklist

## 🔍 Testing Instructions
Ground truth is the Jellyfin server log (`Playback stopped ... Stopped at X ms`) and each item's resume position in the web client.

1. With auto-play next enabled, play episode A in the native player from a few seconds before its end and let the countdown swap to episode B.
2. Make B slow to load (a transcode or a throttled network) so A reaches its end while B is still loading.
3. Exit the player as soon as B starts. Before: B's stop line carries A's end-of-file position. After: B's stop line carries B's real position (a few seconds in).
4. Regression check: play any item to its natural end with no swap. The stop line must still carry the end position and the item must be marked played.

The guard is covered by `utils/nativePlayer/applyPlaybackEnded.test.ts`; the race itself has not been reproduced on a device, hence the unticked platform box.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved playback completion handling so the final position is saved accurately.
  * Prevented stale end-of-playback events from interrupting content while a new stream is loading.
  * Improved stability when playback events occur without an active session.

### DIFF
--- a/providers/NativePlayerProvider.tsx
+++ b/providers/NativePlayerProvider.tsx
@@ -100,6 +100,7 @@ import {
   type SubtitleSelectablePlayer,
 } from "@/utils/jellyfin/subtitleUtils";
 import { logAndCaptureError, writeToLog } from "@/utils/log";
+import { applyPlaybackEnded } from "@/utils/nativePlayer/applyPlaybackEnded";
 import { applyProgressTick } from "@/utils/nativePlayer/applyProgressTick";
 import {
   buildNativePlayerConfig,
@@ -1640,9 +1641,7 @@ const NativePlayerProviderInner: React.FC<{
       }),
 
       addNativePlayerListener("onPlaybackEnded", (payload) => {
-        const session = sessionRef.current;
-        if (!session) return;
-        session.positionMs = payload.positionSec * 1000;
+        applyPlaybackEnded(sessionRef.current, payload.positionSec);
       }),
 
       addNativePlayerListener("onDismiss", (payload) => {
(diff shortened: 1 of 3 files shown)


fredrikburmester:
Closing: review showed the guard does not change the reported position. Both native EOF handlers emit onPlaybackEnded and then dismiss in the same call, and the dismiss carries the same outgoing-stream position into teardownSession, which takes the later of the tracked and native values. The fix has to live in the teardown path for a session that never loaded; a replacement PR follows.

PR #2021, opened 2026-08-31
fix(settings): make the intro crash-report toggle disableable

lancechant:
# 📦 Pull Request

[AI Assisted](#)

## 📝 Description

Tapping the crash-report switch on the intro sheet could not turn it off. Three fixes, one user-visible bug:

- **Tap path (the reported repro)**: the row press is the single mutation path and the switch was made presentational with `pointerEvents="none"` on the Switch itself — but a Switch ignores its own `pointerEvents` on Android, so its native touch handler flipped it and the controlled value snapped it straight back. The switch is now wrapped in a `View pointerEvents="none"`, which a parent View honors, so a tap anywhere on the row — switch included — falls through to the row.
- **Plugin sync reverting consent**: the login-time plugin sync seeded the plugin's unlocked `sentryEnabled` default over an opt-out made on the pre-login intro sheet (the applied-defaults record predates the setting). Consent keys are now never seeded via `NEVER_SEED_KEYS` in the plugin-settings policy. Admin enforcement via locks is unchanged.
- **In-flight refresh race**: a plugin refresh merged from the hook's render snapshot, resurrecting a toggle that landed while its fetch was in flight. The write now evaluates the new pure `pluginRefreshOverlay` against the atom's value at apply time.

An admin-locked toggle now renders as locked ("Disabled by admin") on the intro, mobile settings and TV settings instead of silently eating the press.

## 🏷️ Ticket / Issue

N/A — found during internal testing.

N/A — no visual change when unlocked; locked state reuses the existing "Disabled by admin" pattern.

## ✅ Checklist

## 🔍 Testing Instructions

1. Fresh install (or set Settings → Intro → Reset intro) so the intro sheet shows
2. On the intro sheet, tap the crash-report **switch itself** — it now toggles off (previously it snapped back on Android); the whole row also still toggles
3. Sign in to a Jellyfin server with the Streamyfin plugin — the toggle stays off (previously the login-time plugin sync could flip it back on)
4. Settings → Plugins → Crash reports and the TV settings equivalent behave the same
5. Optional, admin-locked case: with the plugin config locking `sentryEnabled`, all three surfaces show "Disabled by admin" and don't accept presses

## Summary by CodeRabbit

* **New Features**
  * Added administrator controls for crash-reporting settings, including disabled controls and explanatory warnings when locked.
  * Improved TV settings behavior for administrator-locked crash-reporting options.

* **Bug Fixes**
  * Prevented crash-report consent from being enabled automatically during plugin settings refreshes.
  * Improved settings refresh reliability and preserved current user settings.
  * Fixed interaction handling for disabled switches in introductory settings screens.

* **Documentation**
  * Documented switch interaction behavior and the recommended accessibility pattern.

### DIFF
--- a/.claude/learned-facts/switch-pointerevents-ignored.md
+++ b/.claude/learned-facts/switch-pointerevents-ignored.md
@@ -1,0 +1,25 @@
+# Switch Ignores Its Own pointerEvents
+
+**Date**: 2026-08-31
+**Category**: ui
+**Key files**: `components/IntroSheet.tsx`, `components/common/SettingSwitch.tsx`
+
+## Detail
+
+`<Switch pointerEvents="none">` does not make a React Native Switch inert on
+Android (and is unreliable on iOS): the native control consumes its own taps,
+flips visually, and the controlled `value` prop snaps it straight back. The
+user sees a toggle that can be tapped but never changes.
+
+Make a presentational switch inert by wrapping it in a `View
+pointerEvents="none"` — a parent View does honor the prop, so the tap falls
+through to the wrapping pressable (IntroSheet's crash-report row relies on
+this: the row's TouchableOpacity is the single mutation path). Setting the
+prop directly on the Switch is why the intro crash-report toggle ignored taps
+on the switch itself while taps on the row text worked.
+
+## Symptom pattern
+
+One switch "can't be toggled" while every other switch in the app works. The
+working ones wire `onValueChange` directly; the broken one relies on
+`pointerEvents` on the Switch plus a row-level press handler.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -53,6 +53,7 @@ UI and headers:
 - `macos-header-buttons-fix` | macOS Catalyst: use RNGH Pressable, not RN TouchableOpacity
 - `header-button-locations` | Defined in _layout.tsx, HeaderBackButton, Chromecast, RoundButton, etc.
 - `stack-screen-header-configuration` | Sub-pages need explicit Stack.Screen with headerTransparent + back button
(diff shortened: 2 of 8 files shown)


Gauvino:
@lancechant you did test it ? on most device ?

lancechant:
> @lancechant you did test it ? on most device ?

Yup tested on a real android device, why?

Gauvino:
> > @lancechant you did test it ? on most device ?
> 
> Yup tested on a real android device, why?

Didn't test on iOS then ?

lancechant:
> > > @lancechant you did test it ? on most device ?
> > 
> > 
> > Yup tested on a real android device, why?
> 
> Didn't test on iOS then ?

Don't have an IOS device yeah, but should be the same, I hope, if you can test that would be great

Gauvino:
> > > > @lancechant you did test it ? on most device ?
> > > 
> > > 
> > > Yup tested on a real android device, why?
> > 
> > 
> > Didn't test on iOS then ?
> 
> Don't have an IOS device yeah, but should be the same, I hope, if you can test that would be great

Will do for sure, if you want you can ask when not sure if you want someone to test on iOS or Apple TV since you're the TV guy! 🤣

lancechant:
> > > > > @lancechant you did test it ? on most device ?
> > > > 
> > > > 
> > > > Yup tested on a real android device, why?
> > > 
> > > 
> > > Didn't test on iOS then ?
> > 
> > 
> > Don't have an IOS device yeah, but should be the same, I hope, if you can test that would be great
> 
> Will do for sure, if you want you can ask when not sure if you want someone to test on iOS or Apple TV since you're the TV guy! 🤣

Cool thanks, got mobile and tv for android so can test that for any future work 😎

 I did assume this would work, so hoping that is still the case😅

Gauvino (approved):
Tested on iOS work's like expected

PR #1265, opened 2025-12-08
feat: add autorotate for landscape

Simon-Eklundh:
# 📦 Pull Request

## 🛠️ What’s Changed

- Type: feat | fix | docs | style | refactor | perf | test | chore | build | ci | revert
- Scope (optional): e.g., auth, billing, mobile
- Short summary: what changed and why (1–2 lines)
-->

## ✅ Checklist

- [x] I’ve read the [contribution guidelines](CONTRIBUTING.md)
- [x] Code follows project style and passes lint/format (`npm|pnpm|yarn|bun` scripts)
- [x] Type checks pass (tsc/biome/etc.)
- [x] Docs updated (README/ADR/usage/API)
- [x] No secrets/credentials included; env vars documented
- [x] Release notes/CHANGELOG entry added (if applicable)
- [x] Verified locally that changes behave as expected

PR #1178, opened 2025-11-13
refactor: better settings

fredrikburmester:
Refactor settings page with grouping.

PR #2070, opened 2026-09-15
fix(settings): offer landscape auto in the other settings orientation picker

### LINKED ISSUE #110: Video orientation defualt missing option for "Landscape Auto"
For the video orientation option, the iOS app has an option for "Landscape Auto". However, the defaults in this plugin only contains options for "Landscape Left" and "Landscape Right". Please add in "Landscape Auto" as well!

Gauvino:
[AI Assisted](#)

## 📝 Description

The orientation list is built by hand in two places and the two had drifted apart.

| File | Offers `LANDSCAPE` |
|---|---|
| `components/settings/PlaybackControlsSettings.tsx` | yes |
| `components/settings/OtherSettings.tsx` | no |

Both read the same `home.settings.other.orientations.*` keys, and
`"LANDSCAPE": "Landscape auto"` has been in `translations/en.json` the whole time. One
list was updated and the other was not, so the option simply never appeared on the Other
settings screen.

The two blocks are byte identical after this change. Nothing else moved: no new string,
no new value, no behaviour beyond the option becoming selectable.

Worth knowing: the value was already honoured. `defaultVideoOrientation` is read
wherever it is set, so a server that pushed `Landscape` through the Streamyfin plugin
already worked. What was missing was a user being able to choose it for themselves.

## 🏷️ Ticket / Issue

Fixes streamyfin/jellyfin-plugin-streamyfin#110, reported by two people.

The plugin side of that issue is streamyfin/jellyfin-plugin-streamyfin#167, which labels
the same value "Landscape auto" in the admin form so the two screens stop naming one
choice two ways.

The second request on that thread, opening video expanded rather than cropped by
default, is not this and is not addressed here.

## 🖼️ Screenshots / GIFs (if UI)

N/A, the row renders exactly like the four already in that list.

## ✅ Checklist

- [x] I have read the contribution guidelines
- [ ] I have verified this works on all target platforms
- [x] Lint and typecheck considerations: the change adds an enum member the sibling
      component already uses and a key to an inferred object literal, so it is type safe
      by construction. **A full typecheck was not run here**, see below.
- [x] No secrets or credentials
- [x] AI assistance is declared

## 🔍 Testing Instructions

Not yet tested on a device, and that is the box left unticked above. What to check:

1. Settings, then Other settings, then the video orientation row
2. **Landscape auto** appears between Portrait up and Landscape left
3. Selecting it persists, and the row's summary reads "Landscape auto" afterwards
4. Playing a video with it selected rotates freely between both landscape directions
   rather than locking to one
5. Playback controls settings, where the option already existed, still lists the same
   five and behaves the same

Worth doing on both iOS and Android, since orientation locking is the one part of this
that is not shared.

## Summary by CodeRabbit

* **New Features**
  * Added landscape as an available video orientation option in settings.
  * Added the corresponding translated label for the new option.


Gauvino:
Closing this, it was wrong. OtherSettings.tsx is dead code. Nothing imports it, there is no dynamic import either, and the last commit that mounted it under app/ was #1178. So this edits a file that has not rendered in months.

The screen that actually renders is PlaybackControlsSettings, from settings/playback-controls/page.tsx, and it has offered LANDSCAPE since #1265 in January, seven months before the issue was filed.

I diffed the two components and assumed both were live without checking that either was mounted. That was the mistake.

Re-reading the issue body settles it anyway. It says the iOS app has the Landscape Auto option and the plugin defaults only have left and right, so it was never an app bug. The plugin side is streamyfin/jellyfin-plugin-streamyfin#167.

PR #2081, opened 2026-09-18
feat(seerr): generate the API types from the pinned OpenAPI spec

Gauvino:
# 📦 Pull Request

[AI Assisted](#)

## 📝 Description

First part of taking the Seerr types from the published OpenAPI spec instead of `utils/jellyseerr`, the git submodule pointing at a fork of the server. Nothing consumes what this adds yet, so nothing in the app changes.

`scripts/seerr/generate-types.ts` fetches `seerr-api.yml` at the tag named in `utils/seerr/generated/pin.json`, writes the types, and records the source sha256 in both the pin and the generated file's header, so an edit by hand shows up. `bun run seerr:types` regenerates everything.

It also writes `api-shapes.json`, the property paths each response declares, flattened. That is what the contract test in a later part will compare against a real server, and it is what makes a spec bump reviewable: a route that appears or a property that moves is one line there, against a hundred in the generated types.

Two shapes needed care, and both were found by running the flattening against the real spec rather than by reading it:

- a union collapses to the union of its variants, because `/search` and `/discover/trending` return one of three result shapes and a caller gets all three. Without it, neither route described a single result property;
- a map whose keys are data, like the certifications keyed by country code, is described once as `*`. Listing its keys would describe the server's content, and it made 104 country codes read as undeclared fields.

`.github/workflows/seerr-spec.yml` reads the latest Seerr release every Monday and, when it differs from the pin, regenerates and opens a pull request. Nothing merges on its own. The tag comes from a repository we do not control and ends up in a branch name and a commit message, so the job refuses anything that does not look like a version.

`utils/seerr/generated` is excluded from Biome the same way `utils/jellyseerr` and `docs/jellyfin-openapi-stable.json` already are, so the generated file stays byte for byte what the generator produces.

The design and the plan for the rest are in `docs/superpowers/specs/2026-09-18-seerr-migration-design.md` and `docs/superpowers/plans/2026-09-18-seerr-migration.md`, on the base branch.

## 🏷️ Ticket / Issue

Part of the jellyseerr to seerr migration. Targets `seerr-migration`, not `develop`.

N/A, no UI. Nothing imports the generated types yet.

## ✅ Checklist

## 🔍 Testing Instructions

1. `bun install`, then `bun run seerr:types`. It prints `seerr-team/seerr@v3.4.1: 10227 lines` and leaves the working tree clean: what is committed is exactly what the generator produces.
2. `bun test scripts/seerr/` covers the flattening: nested objects and arrays, a schema that refers to itself, `allOf`, unions, free-form maps, and an operation that answers without a JSON body.
3. `bun run typecheck`, `bun run lint`, `bun run format` and `bun run i18n:check` all pass. `bun run doctor` fails on 34 out-of-date Expo packages, which it already does on `develop`.
4. The flattening was also checked against reality, not only against its own tests: comparing `api-shapes.json` with response shapes captured from a running Seerr reports the same 12 routes as an independent measurement made with a separate script, and the small difference in field counts is the gap between `v3.4.1` and the `develop` the measured server runs.
5. The workflow itself cannot be proven by a green job. Run it by hand from the Actions tab once this is merged, with the pin moved back a version, and check that the pull request it opens carries both the regenerated types and the moved pin.

## Summary by CodeRabbit

* **New Features**
  * Added automated weekly and on-demand checks for new Seerr API releases.
  * Added automatic generation and validation of Seerr API types and response shape metadata.
  * Added safeguards to verify downloaded API specifications before regeneration.
  * Added automated pull request creation and optional squash auto-merge for updated API definitions.

* **Tests**
  * Added coverage for specification pinning, integrity checks, schema traversal, arrays, maps, unions, and referenced schemas.


Gauvino:
@coderabbitai review

Gauvino:
Good catch, fixed in d77f0453. The job installs dependencies with `contents: write`, so the persisted token was reachable from an install script; `create-pull-request` takes its own token and the `gh` steps read `GH_TOKEN`, so nothing here needed it.

PR #2086, opened 2026-09-21
feat(seerr): declare the values and the shapes the spec cannot give

Gauvino:
# 📦 Pull Request

[AI Assisted](#)

## 📝 Description

Second part of taking the Seerr types from the published spec instead of the `utils/jellyseerr` submodule. Still nothing consumes it: the app keeps reading the submodule until the next part switches it over, so nothing in the app changes.

Three files, and each one exists for a reason the generator cannot cover.

**`permissions.ts`, and the enums in `types.ts`.** The spec serves `permissions`, `status`, `issueType` and the rest as a plain `number`, with what the numbers mean written in a description. Generation gives `status?: number`, which is worse than what the submodule gave us, so these are ported from upstream and their values are asserted in tests: a wrong one reads as a different state and nothing else would catch it. The only difference from the copy the app reads today is a rename upstream did, `BLACKLIST` to `BLOCKLIST`, same values, and the app reads neither.

**`corrections.ts`.** Fourteen routes where a measured server disagrees with the spec, each with the date it was measured and a line saying what happens. `types.ts` encodes the same gaps as real types, and the contract test in the next part reads the list. One test already refuses a correction the spec covers, which is how we will learn the day upstream fixes one, and it caught one of mine that had never been real.

**`data.ts`.** The networks, studios and genre colours the Seerr web interface owns. Nothing serves them over HTTP, so no spec can describe them. Copied and then checked value by value against what the app reads today: 22 networks, 10 studios, 28 genres, identical.

One design choice worth naming: the union of search results is discriminated on `mediaType` here. The spec types that field as a plain string, so a caller cannot tell a movie from a series without it, and every screen that draws a result needs to.

## 🏷️ Ticket / Issue

Part of the jellyseerr to seerr migration. Targets `seerr-migration`, not `develop`. Builds on #2081.

N/A, no UI. Nothing imports any of this yet.

## ✅ Checklist

## 🔍 Testing Instructions

1. `bun test utils/seerr/` is 26 tests across the four files.
2. `bun run typecheck` is part of the test here rather than a formality. The types derived from the generated spec resolve to `never` when the shape they reach for has moved, and `never` accepts nothing, so `utils/seerr/types.test.ts` assigns a literal to each one. It failed twice while this was written, which is what it is for.
3. `bun run lint`, `bun run format` and `bun run i18n:check` pass. `bun run doctor` fails on 34 out-of-date Expo packages, which it already does on `develop`.
4. The copied tables were compared against the submodule rather than eyeballed: `genreColorMap`, `colorTones`, `networks`, `studios` and `COMPANY_LOGO_IMAGE_FILTER` all serialise identically to what the app reads today.
5. The corrections come from response shapes captured on a running Seerr, not from reading the spec. The next part adds the test that keeps them honest.

## Summary by CodeRabbit

- **New Features**
  - Added Seerr-compatible media, user, request, issue, service, and discovery data support.
  - Added permission checks for administrator and granular access levels.
  - Added support for network, studio, genre, color, and media metadata.
  - Improved handling of live server responses where they differ from the published API specification.

- **Bug Fixes**
  - Corrected the documented season-route discrepancy and clarified external ID handling.

- **Tests**
  - Added comprehensive validation for permissions, data mappings, response types, and API corrections.


Gauvino:
@coderabbitai full review

Gauvino:
All three applied in db3096cc.

The A24 one was the good catch: the studios came from the submodule, and the submodule had lost A24. The whole point of this migration is not to inherit that fork's staleness, and the first thing I did was inherit it. Checked every row against the pinned version this time, and that was the only difference. The shape stays the submodule's, an id and a bare image path, because that is what the app's components read.

`nullable` and `absent` are now held against the declared paths the same way `added` already was.

`hasPermission` drops the mutable accumulator. Since that file is a port, I checked the rewrite against upstream's own version over 2508 combinations of asked permissions, held permissions and check types: no difference.

PR #2088, opened 2026-09-21
test(seerr): check the types against what a real server sends

Gauvino:
# 📦 Pull Request

[AI Assisted](#)

## 📝 Description

Third part of taking the Seerr types from the published spec. Stacked on #2086, so it targets that branch and will retarget itself to `seerr-migration` when #2086 merges. Nothing in the app consumes any of this yet.

Seerr validates requests against its spec and nothing on the way back, so the response half is documentation, and it has drifted. This measures the drift and keeps measuring it.

**`scripts/seerr/capture.ts`** calls the routes the app reads and writes their shape under `utils/seerr/__fixtures__`: keys and the type of each value, never a value. The fixtures live in the repository and are read in review, so a captured title or token would be a leak rather than a test. A test in the suite refuses any string in a fixture that is not a type name.

**`utils/seerr/contract.test.ts`** puts three things side by side, with no network and no secret: what a server sent, what the spec declares, and what `corrections.ts` says about the difference. It fails on a property nobody knows about, which is upstream moving, and on a correction the spec has caught up with, which is our layer carrying an entry it no longer needs.

Its first run found **34 gaps on four routes nobody had measured**, all real:

- `MediaInfo` declares 6 of the 25 properties it carries, on both `/movie/{movieId}` and `/tv/{tvId}`;
- `watchProviders` is declared as an array of arrays and sent as one array;
- `ProductionCompany`, `Network` and the series' `createdBy` are each missing a few fields.

They are in `corrections.ts`, dated today.

Two of my own mistakes came out of the same run, which is the point of the thing: a correction only applies where the capture actually went, because a film that is not in the library carries no `mediaInfo` at all, and the rename check only asserts the declared side, because `watchProviders[]` is both what the server sends and what the spec uses for the outer level.

## 🏷️ Ticket / Issue

Part of the jellyseerr to seerr migration. Builds on #2081 and #2086.

N/A, no UI.

## ✅ Checklist

## 🔍 Testing Instructions

1. `bun test utils/seerr/ scripts/seerr/` is 89 tests, all green, and needs no server.
2. Check that the contract test can fail, because one that has never been seen red proves nothing. Add a line to any fixture by hand, say `"somethingSeerrAdded": "string"` in `utils/seerr/__fixtures__/status.json`, run it again. It fails and names the field. Take it back out.
3. Check the fixtures carry no values: `grep -o ':"[^"]*"' utils/seerr/__fixtures__/*.json | sort -u` returns only type names.
4. To measure against your own server: `SEERR_URL=... SEERR_JELLYFIN_USER=... SEERR_JELLYFIN_PASSWORD=... bun run seerr:capture`. It signs in the way the app does and rewrites the fixtures. Read the diff before committing it.
5. `bun run typecheck`, `lint`, `format` and `i18n:check` pass. `doctor` fails on out-of-date Expo packages, as it already does on `develop`.

## Summary by CodeRabbit

* **New Features**
  * Added tooling to capture Seerr API response shapes from a configured instance.
  * Added broad response-shape coverage for discovery, search, media, users, requests, services, settings, and status endpoints.
  * Added read-only route coverage for safer API inspection.

* **Tests**
  * Added automated contract checks comparing observed responses with declared API specifications.
  * Added validation for response shapes, nested paths, arrays, null values, and route corrections.


Gauvino:
@coderabbitai full review

Gauvino:
@coderabbitai review

Gauvino:
All five applied in 3b99eccf, and two of them were holes in the measurement rather than in its plumbing.

The first-element one was the important catch. A shape came from `value[0]`, so a property one film carries and another does not was missing from the fixture and the contract test passed on a response it had never seen whole. Every element is merged now. Keeping the nullability cost a second pass: merging a `null` with a value and keeping the value threw away one of the three kinds of gap this exists to find, so a leaf that disagrees with itself now reads `null|string`.

Depth is seven rather than three, which reaches `mediaInfo.requests[].requestedBy.id`.

Each route names the status it expects and anything else stops the capture, so a 401 can no longer replace a real check with a fixture that has no shape.

The leak check walks the shape instead of matching quoted text, and the contract test validates nullable, absent and renamed entries rather than only the added ones.

Measuring whole responses turned up what the first pass had missed: MediaInfo is under-declared everywhere it appears, nested in a film, a series, a search result and a credit alike. Listing its 19 properties at every depth would have said the same thing fifty times, so corrections gained a shallow kind for a schema the spec declares without describing its contents.

